PR #39, opened 2025-08-20
#37 - Add range-read support

jallum:
The Basalt storage driver isn't really as friendly to range-reads as I'd like. This is a sketch for a new storage driver, Olivine.

Olivine is a multi-version concurrency control (MVCC) storage driver for Bedrock that implements a page-based B+ tree-like structure with sliding window version management over a shard. By using a hybrid tree/page structure, the aim is to be able to quickly find a page or run of pages that can be passed efficiently to another process for extraction/loading from external storage. The tree structure is designed to require minimal changes during mutation (only when pages split or are appended-to). The use of an indirection-layer (page map) allows maximum sharing of structure between versions inexpensively. 

This hybrid structure allows Olivine to maintain a _complete_ in-memory index of the shard _for each version_ in the MVCC window, making operations like range-clear and range-scan _very_ fast and simple.

## Key Architectural Concepts:

1. Page-Based Storage
- [x] Data is organized into pages (max 256 keys each) that form a linked chain
- [x] Each page contains sorted keys and their corresponding versions
- [x] Pages are to be encoded as binary data with a 16-byte header for efficient sharing between processes
- [x] Uses simple median split algorithm when pages exceed capacity for now

2. Multi-Version Concurrency Control (MVCC)
- [x] Lookaside buffer (ETS table) stores recent key-value pairs for fast access
- [x] Sliding time window (5-second default) retains multiple versions in memory
- [x] Version filtering allows point-in-time queries within the window
- [x] Older versions are automatically evicted and persisted to DETS

3. Tree-Based Index
- [x] Interval tree (GB-trees) maps key ranges to page IDs for O(log n) lookups
- [x] Tree stores {last_key → {page_id, first_key}} mappings
- [x] Supports efficient range queries by walking the tree for page ids

4. Hybrid Memory/Disk Architecture
- [x] Hot data: Recent versions stay in memory (lookaside buffer)
- [x] Warm data: Page metadata cached in memory, content loaded on-demand
- [x] Cold data: Older versions persisted to DETS for durability

5. Transaction Processing
- [x] Applies mutations (set/clear/clear_range) to create new versions
- [x] Page splitting handles growth automatically
- [x] Page modifications are minimized by pushing mutations out to the pages in one pass, and applying them in another.
- [x] Batch persistence for efficient disk writes

### Point Reads

The general idea is to keep the main version structure accesses as _quick_ as possible, and to hand off most of the grunt work to another process to finish the job, and return the value to the caller. Because of the shared-nothing philosophy of the BEAM, even though things are immutable, they're still _copied_ from one process heap to the other during a message send. Reading from an :ets table doesn't get around this (as things are just term_to_binary'd under the hood), so efficient sharing is only really possible through the passing of binary-values[^1].

Here's a sketch of the technique:

1. Receive a {:fetch, read_version, key} message
2. Walk back through the MVCC versions, until a suitable {tree, page_map} is found for the given read_version.
3. Use the tree to find the page-id that could contain the key, return the page from the page_map.
4. Spin up a task to:
   - search the binary page for the last write-version of the key, if it exists.
   - if the version is within the MVCC window (>= durable_version), the {version, key} _must_ exist in the LookasideBuffer, fetch it from there.
   - otherwise, the key _must_ exist in unordered k/v storage (:dets), fetch it from there.
   - reply to the caller with the fetched value.

### Range-Reads

This idea is extended when it comes to range reads. The tree structure is only necessary to lookup the possible starting page of a key. From there, the page map and the next_id pointers in each page allow the pages that contain the range to be quickly identified and returned in-order. The flow for a range read is similar to a point-read, but the gains are _much_ larger, since the only thing that's being passed to the worker task (that needs to be _copied_) is the _list_ of the binaries. All of the lookup / disk fetching work can be offloaded to another process -- allowing the next request to be read and other lookups to start.

1. Receive a {:fetch, read_version, start_key, end_key} message
2. Walk back through the MVCC versions, until a suitable {tree, page_map} is found for the given read_version.
3. Use the tree to find the page-id that could contain the key, and follow the page chain until end_key >= the first key of the page. return the a list of pages from the obtained from the page_map.
4. Spin up a task to:
   - search the binary page for the last write-version of the first key, if it exists.
   - if the version is within the MVCC window (>= durable_version), the {version, key} _must_ exist in the LookasideBuffer, fetch it from there.
   - otherwise, the key _must_ exist in unordered k/v storage (:dets), fetch it from there.
   - emit the {key, value} pair
   - repeat across this and subsequent pages, until end_key is reached.
   - reply to the caller with the fetched keys and values.

## Todo list:
- [x] Add a log puller process for getting new transactions
- [x] expose range read to the `Repo` interface
  - [x] Double-check that range reads are properly accounted for in TransactionBuilder. 
- [x] Create an example of range reads / clears in action.

[^1]: binary values >64 bytes are the _only_ BEAM types that are not _copied_ between processes.


lib/bedrock/data_plane/storage/olivine/database.ex
```diff
@@ -0,0 +1,200 @@
...
+  @moduledoc false
+
+  @type page_id :: non_neg_integer()
+
+  @opaque t :: %__MODULE__{
+            dets_storage: :dets.tab_name(),
+            window_size_in_microseconds: pos_integer()
+          }
+  defstruct dets_storage: nil,
+            window_size_in_microseconds: 5_000_000
```
jessestimpson, line 11: Any future interest in external configurability of the window size? It's a common request from FDB users, but to be honest the 5s constraint does have positive effects on encouraging good design of data models, IMO. So I'm of two minds about it.

lib/bedrock/data_plane/storage/olivine/version_manager/page.ex
```diff
@@ -0,0 +1,290 @@
...
+    last_key_offset = if key_count > 0, do: 32 + key_count * 8 + byte_size(encoded_keys) - 2, else: 0
+
+    encoded_versions = for version <- page.versions, into: <<>>, do: version
+
+    header = <<
+      page.id::unsigned-big-64,
+      page.next_id::unsigned-big-64,
+      key_count::unsigned-big-32,
+      last_key_offset::unsigned-big-32,
+      0::unsigned-big-64
```
jessestimpson, line 53: Is this 0 a placeholder for something?

lib/bedrock/data_plane/storage/olivine/version_manager/page.ex
```diff
@@ -0,0 +1,290 @@
...
+      nil -> {:error, :not_found}
+      index -> {:ok, Enum.at(page.versions, index)}
+    end
+  end
+
+  @spec add_key_to_page(page :: page(), key :: binary(), version :: Bedrock.version()) :: page()
+  def add_key_to_page(page, key, version) when is_binary(version) and byte_size(version) == 8 do
+    case find_insertion_point(page.keys, key) do
+      {index, :duplicate} ->
+        new_versions = List.replace_at(page.versions, index, version)
```
jessestimpson, line 115: How do you find the performance of these List operations? If I change a key 10000 times in the window, I'm gathering that the list of versions will be 10000 long, and I wonder if these start becoming bottlenecks?

jessestimpson (approved):
Cool! Very exciting to see range progress. I only had limited time to devote, but threw in a few minor questions.

lib/bedrock/data_plane/storage/olivine/version_manager/page.ex
jallum: poor! ;)

the idea is to keep things simple and work out the kinks of the overall idea, and then to dial in the performance once i have a better feel for how it works. i'm just not smart enough to get there in one step. things should improve dramatically when i switch over to using the binary representation and random access is possible.

roll it around, i'm very open to suggestions.

jallum: i was thinking, though, that one interesting approach might be to do a pass where the mutations are distributed out to the pages that they affect, and then they could be applied all at once, per-page... so better cache locality and fewer (maybe just one) list operation per page. for range-clears, though, you'd do those _immediately_, and any unapplied mutations on pages that get cleared would just get blown away with the page... no need to do _anything_.

jallum:
Latest progress on the new MVCC storage driver:
- Fixed critical persistence bottleneck with intelligent deduplication
- Database now eliminates redundant writes using newest-first processing
- Comprehensive test coverage added for all persistence workflows
- Buffer management streamlined for better performance
- Ready for range-read integration next phase

Performance improvements: Reduced DETS file size + eliminated duplicate key persistence through efficient last-writer-wins semantics.

The hybrid tree/page structure with sliding window MVCC is working well - next up is exposing range operations to the Repo interface and pulling from the logs.

PR #40, opened 2025-08-20
NITs I found while reading

hissssst:
This is a small PR with a bunch of unrelated changes and optimizations and I am making these mostly as a place to discuss these changes. I'll make separate PRs if you're ok with the changes

### DIFF
--- a/lib/bedrock/data_plane/commit_proxy/finalization.ex
+++ b/lib/bedrock/data_plane/commit_proxy/finalization.ex
@@ -94,12 +94,14 @@ defmodule Bedrock.DataPlane.CommitProxy.Finalization do
     ]
     defstruct [
       :transactions,
+      :indexed_transactions,
       :commit_version,
       :last_commit_version,
       :storage_teams,
       :logs_by_id,
       resolver_data: [],
       aborted_indices: [],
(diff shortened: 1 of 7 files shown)


lib/bedrock/data_plane/commit_proxy/finalization.ex
```diff
@@ -454,34 +457,20 @@ defmodule Bedrock.DataPlane.CommitProxy.Finalization do
...
+      process_transaction_for_logs(transaction_with_idx, plan.aborted_indices_set, plan.storage_teams, logs_by_id, acc)
+    end)
+    |> case do
       {:ok, mutations_by_log} ->
         result =
           Map.new(mutations_by_log, fn {log_id, mutations_list} ->
             encoded =
               Transaction.encode(%{
-                mutations: Enum.reverse(mutations_list),
+                mutations: Enum.reverse(mutations_list), # Why reverse here?
```
jallum, line 473: because `clear` before `set` is not the same as `set` before `clear`

hissssst: But aren't these mutations conflict-free at this point?

PR #76, opened 2026-05-26
fix materializer recovery for empty log descriptors

ccarvalho-eng:
## Summary

- Treat empty log descriptors as runtime-routed logs when recovery unlocks shard materializers.
- Keep legacy tag-filtered descriptors scoped to the matching shard.
- Add a fresh-cluster regression for Bedrock 0.5-style empty log descriptors.

## Root Cause

Fresh Bedrock 0.5 layouts use empty log descriptors because shard-to-log routing is computed at runtime. Recovery still filtered materializer unlock logs by shard tag, so `[]` matched no shard and each fresh-cluster materializer started with no logs to pull from.

```mermaid
flowchart LR
    A[Recovery logs] --> B{Descriptor}
    B -->|empty list| C[Runtime routed log]
    B -->|tag list| D[Shard tag filter]
    C --> E[Materializer unlock logs]
    D --> E
    E --> F[Pull committed transactions]
```

Refs bedrock-kv/job_queue#6.


mikehostetler:
LGTM

PR #82, opened 2026-07-14
Index every shard so uncovered keys fail loudly instead of vanishing; fix corrupt-tree bug; add LayoutIndex tests (bedrock-5ap, bedrock-tn5, bedrock-dwu, bedrock-q67.1)

jallum:
## Why we did this

The LayoutIndex is the map the client uses to route reads: "key `banana` lives in shard `a`–`m`, ask that storage server." It had two quiet problems.

First, shards with no storage server were silently dropped from the index. If the shard covering `d`–`h` had no server, the index pretended `d`–`h` didn't exist — those keys became invisible, and navigation "helpfully" skipped over them, which could silently drop live keys during cross-shard reads.

Second, on the *production default* layout (two shards touching at a shared boundary), the index builder created a broken zero-width segment and fed duplicate keys into the underlying `gb_trees` structure, whose behavior is undefined with duplicates (bedrock-tn5). It only appeared harmless because the one caller happened not to hit the corrupt part.

This module also had almost no tests (82.2% line coverage, zero behavioral tests).

## What it does

- **Every shard is now indexed, even server-less ones** (bedrock-q67.1). A shard with no server becomes a segment with an empty server list. Lookups return it, callers fail loudly with `layout_lookup_failed` instead of silently missing data, and the system heals coverage in the background.
- **Fixes the corrupt-tree bug** with boundary dedup, plus a defensive zero-width-segment guard (bedrock-tn5).
- **Makes next/previous segment navigation symmetric** (bedrock-dwu, resolved via bedrock-q67.1): both return the exactly-adjacent segment, including uncovered ones.
- **Adds 31 tests** for LayoutIndex and StorageRacing (coverage 82.2% → 98.2%) (bedrock-5ap).

## How it works

Take three touching shards where the middle one has no server:

```elixir
shard_layout: %{"d" => {1, "a"}, "h" => {3, "d"}, "m" => {2, "h"}}
shard_materializers: %{1 => p1, 2 => p2}   # tag 3 has no server
```

Before, the index only contained `{a,d} → [p1]` and `{h,m} → [p2]`; a lookup for `"e"` crashed with "No segment found", and stepping forward from `"b"` jumped straight to `{h,m}` — or, per the dwu bug, stopped entirely. Now the index is `{a,d} → [p1]`, `{d,h} → []`, `{h,m} → [p2]`: looking up `"e"` returns `{{"d","h"}, []}`, StorageRacing turns that into a loud `layout_lookup_failed`, and `get_next_segment` from `"b"` returns the uncovered `{d,h}` rather than skipping it.

For the tn5 bug: the default layout has shards `<<>>`–`<<0xFF>>` and `<<0xFF>>`–`<<0xFF,0xFF>>`. The shared `<<0xFF>>` boundary appeared twice in the sorted boundary list, creating a zero-width `<<0xFF>>`–`<<0xFF>>` segment and duplicate tree keys. `Enum.dedup()` on the boundaries fixes it; a regression test pins that this layout builds exactly two segments.

Full suite green; audited that the tests pin the contract, not the old broken shape.

Tickets: bedrock-5ap, bedrock-tn5, bedrock-dwu, bedrock-q67.1

### DIFF
--- a/lib/bedrock/internal/transaction_builder/layout_index.ex
+++ b/lib/bedrock/internal/transaction_builder/layout_index.ex
@@ -178,6 +178,10 @@ defmodule Bedrock.Internal.TransactionBuilder.LayoutIndex do
 
     boundaries
     |> Enum.chunk_every(2, 1, :discard)
+    # Defensive only: dedup above makes boundaries strictly increasing, so no
+    # chunk can be zero-width. Kept as a second guard (bedrock-tn5) because a
+    # zero-width segment would put duplicate keys in the orddict, corrupting
+    # the gb_tree.
     |> Enum.reject(fn [segment_start, segment_end] -> segment_start == segment_end end)
     |> Enum.map(fn [segment_start, segment_end] ->
       covering_pids =
--- a/lib/bedrock/internal/transaction_builder/layout_index.ex
+++ b/lib/bedrock/internal/transaction_builder/layout_index.ex
@@ -156,9 +156,11 @@ defmodule Bedrock.Internal.TransactionBuilder.LayoutIndex do
       ranges
       |> Enum.flat_map(fn {start_key, end_key, _pids} -> [start_key, end_key] end)
       |> Enum.sort()
+      |> Enum.dedup()
 
     boundaries
     |> Enum.chunk_every(2, 1, :discard)
+    |> Enum.reject(fn [segment_start, segment_end] -> segment_start == segment_end end)
     |> Enum.map(fn [segment_start, segment_end] ->
       covering_pids =
         ranges
@@ -220,11 +222,13 @@ defmodule Bedrock.Internal.TransactionBuilder.LayoutIndex do
     find_last_segment_before_boundary(iterator, boundary_key, :start_of_keyspace)
   end
 
-  # Find the first segment that starts at the boundary key
+  # Find the first segment that starts at or after the boundary key, walking
+  # forward across any materializer-less gaps (mirrors get_previous_segment,
+  # which walks back across gaps).
   defp find_first_segment_at_boundary(iterator, boundary_key) do
     case :gb_trees.next(iterator) do
       {tree_end_key, {segment_start, pids}, next_iter} ->
-        if segment_start == boundary_key do
+        if segment_start >= boundary_key do
           {:ok, {{segment_start, tree_end_key}, pids}}
         else
           find_first_segment_at_boundary(next_iter, boundary_key)
--- a/lib/bedrock/internal/transaction_builder/layout_index.ex
+++ b/lib/bedrock/internal/transaction_builder/layout_index.ex
@@ -82,6 +82,11 @@ defmodule Bedrock.Internal.TransactionBuilder.LayoutIndex do
 
   This is useful for cross-shard KeySelector resolution when we need to
   continue processing in the next shard.
+
+  Contract: the adjacent segment is the one starting exactly at the current
+  segment's end key. The index has total coverage (every shard is a segment,
+  possibly with `[]` pids), so an adjacent segment always exists until the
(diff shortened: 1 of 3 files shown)


jallum:
Fixed two audit-confirmed LayoutIndex bugs (commit 2d3df7a4), TDD style:

**bedrock-tn5 — boundary dedup in `create_segments_with_pids`**
Boundaries were sorted but not deduped, so contiguous shard layouts — including the production default `%{<<0xFF>> => {1, <<>>}, <<0xFF,0xFF>> => {0, <<0xFF>>}}` from `initialization_phase.ex` — produced a zero-width artifact segment and duplicate keys in the orddict fed to `:gb_trees.from_orddict/1` (violating its unique-keys precondition).

*Failing test evidence (pre-fix):* the regression test built the production default layout and `lookup_range` over the full keyspace returned **3** segments including the zero-width `{<<0xFF>>, <<0xFF>>}` with a merged (wrong) pid list `[data_pid, metadata_pid]`, instead of the expected 2 segments.

*Fix:* `Enum.dedup()` after the boundary sort, plus a defensive `Enum.reject` of zero-width segments before pid assignment.

**bedrock-dwu — `get_next_segment` gap asymmetry**
`find_first_segment_at_boundary` required `segment_start == boundary`, so `get_next_segment` returned `:end_of_keyspace` at materializer-less gaps even when later segments existed — asymmetric with `get_previous_segment`, which walks back across gaps.

*Failing test evidence (pre-fix):* with shards `%{"d" => {1,"a"}, "m" => {2,"h"}}`, `get_next_segment(index, "b")` returned `:end_of_keyspace` despite segment `{h, m}` existing beyond the gap.

*Fix:* changed the match to `segment_start >= boundary`, making the forward walk cross gaps to the next existing segment, mirroring `get_previous_segment`. Updated the characterization test at line 191 to the fixed behavior and added a bidirectional symmetry regression test.

**Verification:** layout_index_test.exs green at 2 seeds (31 tests), `mix test test/bedrock/internal/` green (450 tests), full `mix test` green (2377 tests; one non-reproducible flake in the first run, clean on two subsequent runs). credo + dialyzer pre-commit hooks passed.

jallum:
**Semantic revision (bedrock-q67.1, commit ab62f98d):** this amendment revises part of the earlier fix on this branch (2d3df7a4) following the coverage-and-distribution design in bedrock-q67 (Phase A).

**What changed**

- `LayoutIndex` now builds a **total-coverage** index: both empty-pid filters are gone, so every shard in `shard_layout` becomes a segment — with `[]` pids when no materializer is known for its tag (including tag 0 when `metadata_materializer` isn't a pid). The tn5 boundary dedup and zero-width-segment rejection are kept.
- `get_next_segment` is **reverted to exact-boundary semantics**, undoing the bedrock-dwu gap-walk. `get_previous_segment` was made exactly symmetric (adjacent segment = the one ending exactly at the current start), and the contract is documented on both.
- New regression tests: `lookup_key!` on an uncovered shard returns `{range, []}`, and `StorageRacing.race_storage_servers` on such a key fires its previously-dead `{_key_range, []}` branch, surfacing `{:failure, %{layout_lookup_failed: []}}`.

**Why**

An uncovered shard is not a gap. `shard_layout` partitions the keyspace, so with the filters removed the index can no longer contain holes — the dwu asymmetry can never manifest. More importantly, gap-walking was the wrong instinct: skipping a shard whose materializer is merely unknown would **silently drop live keys** during cross-shard KeySelector resolution. Under the q67 design, coverage is healed by the system (placeholder materializer + Distributor, Phase A) and never skipped by the client — so the client's job is to return the uncovered segment and fail loudly (`layout_lookup_failed`), which becomes a rare backstop once placeholders fill every slot.

This supersedes the bedrock-dwu resolution.

PR #84, opened 2026-07-14
Add behavioral tests for Basalt Logic fetch waitlist and recovery lock (bedrock-8dc)

jallum:
## Summary
- Adds `basalt/logic_test.exs` with 18 tests bringing `Basalt.Logic` from 82.2% to **100% coverage**: shutdown (asserts the dets table actually closes), lock_for_recovery epoch guard + puller teardown, all five `try_fetch_or_waitlist` branches with exact returns, waitlist notification round-trips via real `GenServer.reply` from-tuples, and the live `apply_and_notify` pull pipeline driven deterministically by a minimal FakeLog (drain-signal ordering, no sleeps)

## Bug found (audit-confirmed, ticket filed)
`unlock_after_recovery` sends `{:transactions_applied, version}` via `send(self(), ...)` **inside the puller Task closure** (logic.ex:70), so the notification goes to the task's own mailbox — the server's `handle_info` for it is dead code. Consequences: waitlisted basalt fetches are never notified (clients hang until their own call timeout), waitlist entries leak in server state, and the task mailbox grows unboundedly. Olivine avoids this by capturing `main_process_pid = self()` before the closure. Two-line fix + a server-level regression test tracked in the follow-up ticket.

Also flagged: `Database.fetch`'s `:key_out_of_range` clause matches a `key_range` field the `%Database{}` struct doesn't have (unreachable with real structs), and basalt never wires up waitlist deadline expiry.

## Test plan
- 18/18 green at multiple seeds; full basalt dir 118/118; full suite (2364 tests, 158 properties) green
- Independently audited: deterministic (FakeLog ordering verified happens-after apply), no defects

Ticket: bedrock-8dc

### DIFF
--- a/lib/bedrock/data_plane/materializer/basalt/logic.ex
+++ b/lib/bedrock/data_plane/materializer/basalt/logic.ex
@@ -65,9 +65,13 @@ defmodule Bedrock.DataPlane.Materializer.Basalt.Logic do
   @spec unlock_after_recovery(State.t(), term(), map()) :: {:ok, State.t()}
   def unlock_after_recovery(t, durable_version, %{logs: logs, services: services}) do
     with :ok <- Database.purge_transactions_newer_than(t.database, durable_version) do
+      # Capture the server's pid before building the closure: the closure runs
+      # inside the puller task, where self() would be the task, not the server.
+      main_process_pid = self()
+
       apply_and_notify_fn = fn encoded_transactions ->
         version = Database.apply_transactions(t.database, encoded_transactions)
-        send(self(), {:transactions_applied, version})
+        send(main_process_pid, {:transactions_applied, version})
         version
       end
 
--- a/test/bedrock/data_plane/materializer/basalt/logic_test.exs
+++ b/test/bedrock/data_plane/materializer/basalt/logic_test.exs
@@ -127,6 +127,31 @@ defmodule Bedrock.DataPlane.Materializer.Basalt.LogicTest do
     end
 
     @tag :tmp_dir
+    test "sends {:transactions_applied, version} to the process that started the puller", %{tmp_dir: tmp_dir} do
+      {_otp_name, state} = start_state(tmp_dir)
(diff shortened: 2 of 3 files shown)


jallum:
## Fix: Basalt puller notified itself instead of the server (bedrock-g1r)

**Bug** (`lib/bedrock/data_plane/materializer/basalt/logic.ex`): `unlock_after_recovery/3` built `apply_and_notify_fn` with `send(self(), {:transactions_applied, version})` inside the closure. The closure runs in the puller `Task` (`Pulling.start_pulling` wraps the pull loop in `Task.async`), so `self()` was the task, not the server. The server's `handle_info({:transactions_applied, _}, _)` — the only caller of `Logic.notify_waiting_fetches/2` — was dead code: waitlisted fetches were never notified (clients hung until their call timeout), `waiting_fetches` entries leaked, and the task mailbox grew unboundedly.

**Fix**: capture `main_process_pid = self()` before building the closure and send to it — mirrors the existing Olivine pattern (`olivine/logic.ex`).

**TDD evidence** — both regression tests were written first and confirmed failing for the expected reason:
- `logic_test.exs` — "sends {:transactions_applied, version} to the process that started the puller": pre-fix, `assert_receive {:transactions_applied, ^version_1}` timed out after 5s (mailbox contained only `:fake_log_drained` messages).
- `server_test.exs` — full waitlist round-trip through the GenServer (waitlisted `{:get, ...}` + pulled transaction containing the key): pre-fix, the client never received `{:ok, "bar"}` (`assert_receive` timed out; the waitlisted call hung).

Post-fix: basalt test dir green at 2 seeds (12345, 98765 — 120 tests, 0 failures each); full suite green (13 doctests, 158 properties, 2366 tests, 0 failures).

**Note**: waitlist deadline *expiry* (also mentioned in bedrock-g1r) is intentionally not wired up here — it's a larger design choice and remains open. The `Database.fetch` key_range clause / MVCC naming issues are ticketed separately as bedrock-r3p.

PR #85, opened 2026-07-14
Add tests for Transaction encoding error paths and compare_and_clear (bedrock-aec)

jallum:
## Summary
- Extends `transaction_test.exs` with 27 tests (44 → 71) for the wire-format codec (coverage 97.5% → 98.5%; remaining 4 lines are unreachable)
- Byte-level corruption tests with exact error atoms: truncated section payloads, overcounted section headers, conflict counts exceeding actual range data, undefined opcodes, truncated extended-length parameters — plus `reassemble_sections` round-trips, `compare_and_clear` encode/decode, read_version/read_conflicts validation raises, and the raising `!` wrappers

## Bug found (audit-verified, ticket filed)
`read_write_conflicts/1` violates its `{:error, reason}` @spec: its callback returns bare error tuples but `iterate_sections_data/5` (transaction.ex:392) only matches `{:ok, _}`/`{:found, _}`, so CRC-valid-but-corrupt conflict payloads raise `CaseClauseError`. Blast radius: the resolver (`transaction_conflicts.ex:70`) has an explicit graceful-degrade branch for `{:error, _}` that is unreachable dead code — a buggy peer's payload crashes the resolver process instead. One-line fix + regression test tracked in the follow-up ticket; the implementer wrote the failing tests, confirmed the crash, and removed them rather than codify broken behavior.

## Test plan
- 71/71 green at multiple seeds; full suite (2379 tests, 158 properties) green
- Independent audit traced 6 of the hand-crafted corrupt binaries byte-by-byte against the wire format — all simulate exactly what their test names claim

Ticket: bedrock-aec

### DIFF
--- a/lib/bedrock/data_plane/transaction.ex
+++ b/lib/bedrock/data_plane/transaction.ex
@@ -396,6 +396,9 @@ defmodule Bedrock.DataPlane.Transaction do
 
       {:found, result} ->
         {:found, result}
+
+      {:error, reason} ->
+        {:error, reason}
     end
   end
 
--- a/test/bedrock/data_plane/resolver/conflict_resolution_test.exs
+++ b/test/bedrock/data_plane/resolver/conflict_resolution_test.exs
@@ -92,6 +92,25 @@ defmodule Bedrock.DataPlane.Resolver.ConflictResolutionTest do
       assert failed_indexes == []
     end
 
+    test "handles CRC-valid transaction with truncated conflict data gracefully" do
+      conflicts = Conflicts.new()
+
+      # Fabricate a transaction whose READ_CONFLICTS section claims 5 ranges
+      # but contains no range data. CRCs remain valid (add_section recomputes
+      # them), so only conflict payload decoding fails. Previously this
+      # crashed the resolver with a CaseClauseError; now it degrades to
+      # a transaction with no conflicts.
+      binary = Transaction.encode(%{mutations: [{:set, "key", "value"}]})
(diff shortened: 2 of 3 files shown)


jallum:
Fixed bedrock-jyl (TDD): `read_write_conflicts/1` raised `CaseClauseError` instead of returning `{:error, :truncated_conflict_data}` when a CRC-valid READ_CONFLICTS/WRITE_CONFLICTS section declared more ranges than its payload contained — its callback returns bare `{:error, reason}`, which `iterate_sections_data/5` did not match.

**TDD evidence:** added two regression tests to the existing "corrupted conflict payloads" describe in `transaction_test.exs` asserting `{:error, :truncated_conflict_data}` for both corrupt read- and write-conflict sections. Both failed pre-fix with `CaseClauseError` at `transaction.ex:392` (`no case clause matching: {:error, :truncated_conflict_data}`).

**Fix (minimal):** added an `{:error, reason} -> {:error, reason}` clause to the callback-result case in `iterate_sections_data/5`. Verified the other callback users are unaffected: `extract_and_process_section` wraps errors as `{:found, {:error, _}}` and already handles `{:error, _}` from iteration; `add_section`'s callback only returns `{:ok, _}` and has an `error -> error` fallthrough.

**Bonus:** added a resolver-level test in `conflict_resolution_test.exs` proving the previously unreachable graceful-degrade branch in `extract_conflicts` (transaction_conflicts.ex) now works — a CRC-valid transaction with truncated conflict data no longer crashes the resolver and resolves as a no-conflict transaction.

**Results:** `transaction_test.exs` green at seeds 0 and 424242 (76 tests); full suite green across 3 runs (2382 tests + 158 properties, 0 failures); credo and dialyzer clean. Commit c86ed8a3.

PR #87, opened 2026-07-14
Add behavioral tests for Finalization resolver and reply edge cases (bedrock-9n6)

jallum:
## Summary
- Adds `sharded_resolution_and_edge_cases_test.exs` (16 tests) for `CommitProxy.Finalization` (coverage 95.3% → 99.5%), using the module's own DI seams (`resolver_fn`, `async_stream_fn`, `batch_log_push_fn`, `sequencer_notify_fn`) — no mocking libraries
- Contracts encoded: empty batches commit 0/0 with empty lists to every resolver; aborted indices from multiple resolvers merge and each aborted client gets `{:error, :aborted}`; any resolver error/exit fails the whole plan and all pending clients are told; cross-shard `clear_range` is split and clamped to shard boundaries (hand-verified against the shard table); atomic mutations route by key incl. phash2 tag fallback; log push failures counted, zero-log-services fails rather than vacuously succeeding

## Bug found (audit-confirmed, ticket filed — client-reachable commit-proxy crash)
`ShardRouter.lookup_shards_with_ranges/3` passes `:"$end_of_table"` to `:ets.prev/2` (shard_router.ex:244) when a range starts at/beyond the last shard boundary → `ArgumentError`. Client-side `Tx.clear_range` only guards start >= end, and finalization runs in a task linked to the commit proxy — so `clear_range(<<0xFF,0xFF>>, <<0xFF,0xFF,0>>)` from one client crashes the proxy and kills every transaction in the batch. The `{:storage_team_coverage_error, ...}` branch is dead code until this is fixed. Audit reproduced it empirically.

## Audit
Independent review verified clamping expectations by hand, DI fake fidelity against real call signatures, and determinism (3 seeds). One fix committed: removed a moduledoc bullet claiming coverage-error tests that can't honestly exist until the ShardRouter fix lands. Two more lib findings ticketed: the log-push exit clause won't match real `Task.async_stream` exit shapes, and the `sequencer_notify_fn` typespec says arity-2 vs arity-4 call sites.

## Test plan
- 16/16 green at 3 seeds; full finalization dir (76 tests) green; full suite (2362 tests, 158 properties) green

Ticket: bedrock-9n6

### DIFF
--- a/lib/bedrock/data_plane/commit_proxy/finalization.ex
+++ b/lib/bedrock/data_plane/commit_proxy/finalization.ex
@@ -77,7 +77,8 @@ defmodule Bedrock.DataPlane.CommitProxy.Finalization do
 
   @type timeout_fn() :: (non_neg_integer() -> non_neg_integer())
 
-  @type sequencer_notify_fn() :: (Sequencer.ref(), Bedrock.version() -> :ok | {:error, term()})
+  @type sequencer_notify_fn() :: (Sequencer.ref(), Bedrock.epoch(), Bedrock.version(), opts :: keyword() ->
+                                    :ok | {:error, term()})
 
   @type resolution_error() ::
           :timeout
@@ -915,6 +916,11 @@ defmodule Bedrock.DataPlane.CommitProxy.Finalization do
 
     required_acknowledgments = map_size(log_services)
 
+    # Task.async_stream is ordered by default, so zipping results with the
+    # input log_ids preserves the log_id association even for {:exit, reason}
+    # results, whose reason carries no log identity.
+    log_ids = Enum.map(log_services, fn {log_id, _service_ref} -> log_id end)
+
     log_services
     |> async_stream_fn.(
       fn {log_id, service_ref} ->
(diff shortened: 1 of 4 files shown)


jallum:
Two audit-confirmed bugs fixed, TDD style (failing test first, then fix):

**bedrock-74j — ShardRouter crash on ranges beyond shard coverage** (b419aef0)
- Failing tests first: `lookup_shards_with_ranges/3` with a start_key >= every shard end_key raised `ArgumentError` from `:ets.prev(table, :"$end_of_table")`; a finalization batch containing `clear_range(<<0xFF,0xFF>>, <<0xFF,0xFF,0>>)` crashed with the same raise (would kill the commit proxy via the linked finalization task).
- Fix: `lookup_shards_with_ranges/3` now returns `[]` when `:ets.next` hits `:"$end_of_table"`. This makes finalization's previously-dead `{:storage_team_coverage_error, ...}` branch real: the hostile batch now returns `{:error, {:storage_team_coverage_error, {start, end}}}` and aborts clients with replies instead of crashing.
- Regression tests: shard_router level (beyond-coverage range returns `[]`) and finalization level (coverage error + client abort). Restored the moduledoc bullet for the coverage-failure tests. Point-lookup semantics (`lookup_shard`, `get_logs_for_key` last-shard fallback) left unchanged.
- **Deliberately NOT in this PR:** commit-proxy ingress key validation (rejecting system-range keys at the door) — larger design decision, tracked as bedrock-rag.

**bedrock-dg8 — log-push exit handling + typespec** (92ec9095)
- Failing test first: injecting the real `Task.async_stream` exit shape `{:exit, :timeout}` via `async_stream_fn` raised `FunctionClauseError` (reducer only matched the tagged `{:exit, {log_id, reason}}` shape).
- Fix: results are zipped with input log_ids (async_stream is ordered), so both exit shapes now count as a log failure attributed to the right log. `sequencer_notify_fn()` typespec corrected from arity-2 to the arity-4 shape all call sites use.

Verification: finalization dir + shard_router tests green at seeds 12345 and 98765; full suite green (2365 tests, 158 properties, 0 failures); dialyzer and credo clean (pre-commit hooks).

PR #88, opened 2026-07-14
Add behavioral tests for Directory partition operations (bedrock-ej0)

jallum:
## Summary
- Adds `directory_partition_test.exs` (28 tests) for `Bedrock.Directory` (coverage 92.5% → 99.1%)
- Instead of canned Mox expectations, tests run directory operations end-to-end against an Agent-backed in-memory store that byte-for-byte mirrors `Bedrock.Repo`'s keyspace pack/get_range semantics — covering partition-scoped ops, path-escape validation (`..`), create_or_open flows, accessors, `ToKeyRange`/`ToKeyspace` protocol impls, and all node-value encoding variants via real round-trips

## Bugs found (audit-confirmed against real modules, tickets filed)
1. **Critical — directory hierarchy scans are blind to descendants against any real store.** `Encoding.Tuple.pack/1` wraps top-level lists in a nested-list tag, so a packed node key (`["app"]`) is NOT a byte-prefix of its child (`["app","users"]`) — diverges at the list-terminator byte. Consequently non-root `Directory.list` returns `[]` regardless of children, `remove` orphans descendants, and `move` misses them. Root works only because `pack([]) = <<>>`. Every existing directory test masks this with canned `get_range` mocks that return descendants unconditionally.
2. **`do_move_recursive` crashes against the real repo before the prefix issue even matters**: it calls `repo.get_range(start_key, end_key)` with two binaries, which the generated repo parses as `(range, opts)` → `FunctionClauseError` in `Keyword.get`. Same shape for `clear_range`.

The new tests are deliberately honest about this: they assert only currently-correct behavior (root list, leaf remove/move) rather than pinning the broken paths.

## Audit
Independent review reproduced the prefix divergence and range exclusion by executing lib code directly (no mocks), traced the production plumbing `do_list` uses, and verified nothing in livebooks/bedrock_ex exercises non-root hierarchy ops (why it went unnoticed). Store fidelity verified against `repo.ex`, with one documented leniency: the mock accepts the two-binary get_range/clear_range shape that the real repo rejects (bug #2).

## Test plan
- 28/28 green at multiple seeds; full suite (2374 tests, 158 properties) green

Ticket: bedrock-ej0

### DIFF
--- a/lib/bedrock/directory.ex
+++ b/lib/bedrock/directory.ex
@@ -107,11 +107,18 @@ defmodule Bedrock.Directory do
 
   defmodule NodeKey do
     @moduledoc false
+    # Packs a directory path by concatenating the tuple encoding of each
+    # element (FDB subspace semantics) rather than encoding the path as a
+    # single nested list. This guarantees that a parent's packed key is a
+    # byte-prefix of every descendant's packed key, which the hierarchy
+    # range scans in list/remove/move rely on. The element terminator byte
+    # (0x00, with 0x00 -> 0x00 0xFF escaping inside elements) ensures that
+    # siblings sharing a name prefix (e.g. "app" vs "apple") do not collide.
     def pack([]), do: <<>>
-    def pack(path) when is_list(path), do: Bedrock.Encoding.Tuple.pack(path)
+    def pack(path) when is_list(path), do: Enum.map_join(path, &Bedrock.Encoding.Tuple.pack/1)
 
     def unpack(<<>>), do: []
-    def unpack(packed), do: Bedrock.Encoding.Tuple.unpack(packed)
+    def unpack(packed), do: Bedrock.Encoding.Tuple.unpack_all(packed)
   end
 
   @type directory :: Node.t() | Partition.t() | Layer.t()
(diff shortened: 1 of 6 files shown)


jallum:
## Bug fixes: directory hierarchy scans + move crash (bedrock-j09, bedrock-esg)

Commit f1f8bb85 fixes two audit-confirmed P1 bugs, TDD-style (regression tests written and confirmed failing first — 7 failures — then fixed).

### bedrock-j09 — directory node keys were not byte-prefixes of descendants

`NodeKey.pack/1` encoded the whole path as one nested list (`0x06...0x00`), so `pack(["app"])` diverged from `pack(["app","users"])` at byte 7 and the prefix-range scans in `do_list`/`do_remove`/`do_move_recursive` never covered the subtree against a real store: non-root `list` returned `[]`, `remove` orphaned descendants, `move` missed them (root only worked because `pack([]) == <<>>`).

Fix: `NodeKey` now concatenates the tuple encoding of each path element (FDB subspace semantics); unpacking uses a new `Bedrock.Encoding.Tuple.unpack_all/1`. `Encoding.Tuple`'s general list semantics are **unchanged** — no other callers affected.

### bedrock-esg — do_move_recursive crashed the real repo

`do_move_recursive` called `repo.get_range(start_key, end_key)` / `repo.clear_range(start_key, end_key)` with two bare binaries; the generated repo parses these as `(range, opts)` and dies in `Keyword` handling. It now passes the `{start_key, end_key}` range tuple. The in-memory test harness's forgiving two-binary `get_range`/`clear_range` stubs were removed so the mock matches the real repo interface (this is what let the bug hide).

### ⚠️ KEY-FORMAT CHANGE (breaking, pre-1.0)

This changes the on-disk directory **node key format** (`\xFE`-prefixed directory metadata keys). Directory metadata written by earlier versions will not be readable and **no migration is provided** — use a fresh store. Directory *content* keys (allocated prefixes) are unaffected.

### Evidence

- New regression tests (real-store harness in `directory_partition_test.exs`): non-root `list`, 3-level nesting, name-prefix sibling non-collision (`app` vs `apple`), `remove` deletes whole subtree, `move` relocates all descendants, and a direct byte-prefix property test on `NodeKey.pack`.
- Pre-fix: 7 failures (empty lists, `FunctionClauseError` at `directory.ex:753`). Post-fix: directory tree green at seeds 1 and 424242 (106 tests), full suite green: **13 doctests, 158 properties, 2380 tests, 0 failures**. Credo + dialyzer clean.
- Collateral: mock-based tests updated faithfully to the new format (`build_directory_key`, `@version_key` in `directory_test_helpers.ex`; range-tuple expectations in `move_test.exs` / `version_management_test.exs` — assertions strengthened, not weakened).

PR #91, opened 2026-07-14
Add Distributor singleton skeleton (bedrock-q67.2)

jallum:
## Summary

Adds the **Distributor** — a control-plane cluster singleton that will own shard-coverage policy (on-demand materializer recruitment, placeholder supervision, materializer-death healing, and later split/merge/rebalancing). It is the analogue of FoundationDB's **Data Distributor**: recruited by the director only *after* recovery completes (FDB: the cluster controller recruits DD once the cluster is accepting commits), deliberately **off the request path** — its death degrades coverage healing only, never reads or writes.

Part of the q67 Phase A (Coverage & Distribution) epic.

## What's here

- `Bedrock.ControlPlane.Distributor` (public API) + `Distributor.Server` (GenServer) + `Distributor.State` + `Distributor.Telemetry`
  - State: cluster, epoch, director ref, shard-layout snapshot, materializer monitors (empty for now), placeholder ref (nil for now)
  - Telemetry: `[:bedrock, :distributor, :started]` / `[:bedrock, :distributor, :stopped]`
- **Post-recovery recruitment**: the director's recovery entry points now flow through `advance_recovery/1` (`try_to_recover` + `maybe_start_distributor`); once state reaches `:running`, the director starts the Distributor via the cluster's dynamic supervisor with the current epoch + TSL `shard_layout` snapshot, registered as `cluster.otp_name(:distributor)`
- **Monitoring + re-recruitment**: the director monitors the Distributor pid; a `:DOWN` for it re-recruits at the current epoch instead of tripping the component-failure shutdown path
- **Epoch guard** (`moveKeysLock` analogue): `State.check_epoch/2` is the guard every future shard-map mutation passes through. Two channels for learning the epoch moved on: `check_epoch/3` (call — a caller with a newer epoch supersedes the distributor) and `notify_epoch_change/2` (cast from the director side). Either way the stale distributor logs and stops with `:normal`. It also monitors its director and stops `:normal` when the director exits
- Tests: server lifecycle + epoch guard unit tests; director-level tests for post-recovery recruitment, death → re-recruitment, and epoch N / N+1 supersession

## Deliberately NOT here yet

- **Recruitment logic** (coverage demand handling, node selection via Foreman, TSL deltas) — bedrock-q67.5
- **Placeholder materializer** (bounded parking, fallback coverage endpoint) — bedrock-q67.3
- Service-directory registration and materializer monitoring wiring — arrive with the recruitment flow

## Testing

[...]

PR #92, opened 2026-07-14
Add post-recovery TSL delta application and broadcast (bedrock-q67.4)

jallum:
## Summary

Makes the TSL updatable after recovery, with propagation to clients (bedrock-q67.4, Phase A of the shard-centric storage epic). Pure mechanism — no Distributor or placeholder (separate tickets).

### What already existed vs what's new

- The Link receiver for `{:tsl_updated, new_tsl}` already existed (`link/server.ex`).
- A sender **also already existed**: `Coordinator.State.Changes.broadcast_tsl_update/2` fires on every `put_transaction_system_layout`, and the Director already hands the recovery-built TSL to the coordinator via `notify_transaction_system_layout`. So `{:tsl_updated, ...}` was sent — but only once, at recovery completion. The TSL was static thereafter.
- New: a post-recovery **delta path** on the Director, and a subscription fix so capability-less Links actually receive broadcasts.

### Delta shape

```elixir
@type tsl_delta :: %{Bedrock.range_tag() => pid() | :remove}
```

Minimal map-merge semantics over `shard_materializers` only: a pid puts/replaces the shard's entry; `:remove` deletes it. Applied atomically per call; the rest of the TSL is untouched.

### Epoch guard

`Director.apply_tsl_delta(director, delta, epoch)` applies only when `epoch` exactly matches the director's own epoch; any other epoch (stale, or from a generation this director doesn't know) is rejected with `{:error, :newer_epoch_exists}`, mirroring the `lock_for_recovery` guards on data-plane services. `{:error, :unavailable}` before recovery has produced a TSL. On success the director updates its retained TSL (it already keeps it in `State.transaction_system_layout` — no restructuring needed) and reuses the existing `notify_transaction_system_layout` handoff to the coordinator, which broadcasts to all subscribers.

### Subscription mechanism

Piggybacked on existing machinery rather than building a new registry: the coordinator already held a monitored `tsl_subscribers` MapSet, but only populated it as a side effect of `register_node_resources` — which Links skip entirely when they have no capabilities (pure clients). Added an explicit, idempotent `Coordinator.subscribe_to_tsl_updates/2` cast; `Link.Discovery.change_coordinator/2` now calls it unconditionally on connect. New subscribers are immediately sent the current TSL if one is known (snapshot-then-updates), so late-connecting links don't wait for the next delta. Dead subscribers are pruned via the existing `:DOWN` handling; `send/2` to a dead pid is a no-op, so a dead link never breaks a broadcast.

### Telemetry

`[:bedrock, :director, :tsl_delta_applied]` with `%{shard_count: n}` measurements and `%{cluster, epoch, delta}` metadata (`Director.Telemetry.trace_tsl_delta_applied/3`).

### Type fix

`Coordinator.ref()` now includes `pid()` — the Director already stores and calls the coordinator by pid; dialyzer flagged the omission once the new specs landed.

## Tests

- `test/bedrock/cluster/tsl_delta_round_trip_test.exs` — real `Link.Server` GenServer + thin harnesses delegating to the real `Coordinator.Server`/`Director.Server` handlers: director applies delta → coordinator broadcasts → Link observes the new TSL via `fetch_transaction_system_layout`. Plus stale-epoch rejection, dead-subscriber resilience, and snapshot-on-subscribe.
- `test/bedrock/control_plane/director/tsl_delta_test.exs` — put/remove/mixed delta semantics, missing `shard_materializers` map, epoch guards, `:unavailable` pre-recovery, telemetry event.
- Coordinator `state_test.exs` — subscriber add/remove idempotence, broadcast delivery, dead-subscriber tolerance.

Coordinator + director + cluster dirs green at seeds 0 and 424242; full suite green (2528 tests, 158 properties); `mix format`, `credo --strict`, and dialyzer clean.

### DIFF
--- a/lib/bedrock/cluster/link/discovery.ex
+++ b/lib/bedrock/cluster/link/discovery.ex
@@ -161,12 +161,22 @@ defmodule Bedrock.Cluster.Link.Discovery do
     t
     |> Map.put(:known_coordinator, coordinator_ref)
     |> monitor_known_coordinator()
+    |> subscribe_to_tsl_updates()
     |> register_node_capabilities()
   end
 
   @spec monitor_known_coordinator(State.t()) :: State.t()
   defp monitor_known_coordinator(t) do
     Process.monitor(t.known_coordinator)
+    t
+  end
+
+  @spec subscribe_to_tsl_updates(State.t()) :: State.t()
+  defp subscribe_to_tsl_updates(t) do
+    # Subscribe for {:tsl_updated, tsl} broadcasts regardless of capabilities.
+    # Nodes registering resources are subscribed implicitly by the coordinator,
+    # but capability-less (pure client) links would otherwise never subscribe.
+    Coordinator.subscribe_to_tsl_updates(t.known_coordinator, self())
     t
   end
 
--- a/lib/bedrock/control_plane/coordinator.ex
+++ b/lib/bedrock/control_plane/coordinator.ex
@@ -45,7 +45,7 @@ defmodule Bedrock.ControlPlane.Coordinator do
   alias Bedrock.ControlPlane.Config
   alias Bedrock.ControlPlane.Config.TransactionSystemLayout
 
-  @type ref :: atom() | {atom(), node()}
+  @type ref :: pid() | atom() | {atom(), node()}
   @typep timeout_in_ms :: Bedrock.timeout_in_ms()
 
   @spec config_key() :: atom()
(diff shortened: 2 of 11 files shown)


jallum:
Superseded by #97, which collapses the Phase A stack (#91 → #92 → #94 → #95) into one linear sequence on current develop, carrying all audit corrections. Branch retained for history.

PR #94, opened 2026-07-14
Add placeholder materializer with bounded parking (bedrock-q67.3)

jallum:
> [!NOTE]
> **Stacked PR** — based on `feature/q67-distributor` (#91, the Distributor skeleton). Merge #91 first; this PR should only be reviewed for its own diff.

## What

The placeholder materializer (bedrock-q67.3): the cluster's fallback coverage endpoint. One process per cluster, started, linked, and restarted-on-death by the Distributor. Its pid is intended to fill every uncovered `shard_materializers` slot so clients can never observe an uncovered shard.

## Design

- **Indistinguishable from a real materializer.** Accepts exactly the GenServer call shapes olivine/basalt handle — `{:get, key | KeySelector, version, opts}` and `{:get_range, start, end, version, opts}` — so `StorageRacing`/`PointReads`/`RangeReads` are unchanged.
- **Park** (`Bedrock.Internal.WaitingList` keyed by shard tag) with budget `min(caller opts[:timeout], hold_ms)` (`hold_ms` default 2,000ms); signals `{:coverage_demand, tag}` to the Distributor, deduped to at most one in-flight demand per tag.
- **Timer-driven expiry** (lesson from bedrock-tvb): a single `Process.send_after` armed for the earliest deadline, rescheduled on every waitlist mutation; expired waiters get `{:error, :unavailable}`.
- **Drain on `{:covered, tag, pid}`:** parked requests are re-issued against the live materializer from one `Task` per request (the placeholder never blocks); the pid is remembered so later stale-layout requests are **forwarded** (staleness shim).
- **Shed on `{:coverage_failed, tag, reason}`:** waiters get `{:error, :unavailable}` and the demand dedupe clears so the next request re-triggers recruitment.
- **Distributor side (minimal — recruitment is bedrock-q67.5):** `{:coverage_demand, tag}` is logged, counted (telemetry), and remembered in `pending_demands`; `deliver_coverage/3` and `fail_coverage/3` are the internal/test seam that relays outcomes to the placeholder. The Distributor traps exits, links the placeholder, restarts it on death, and shuts it down explicitly in `terminate/2` (a `:normal` exit signal would otherwise orphan it).
- `WaitingList`'s key type widened to include `Bedrock.range_tag()`.

## Telemetry

`[:bedrock, :distributor, :placeholder, :parked | :forwarded | :drained | :shed]` and `[:bedrock, :distributor, :coverage_demand]` (emitted by the Distributor on receipt).

## Tests

[...]

PR #95, opened 2026-07-14
Add on-demand materializer recruitment (bedrock-q67.5)

jallum:
## Summary

Closes the Phase A loop (bedrock-q67.5): coverage demand → placement → foreman worker start → epoch lock/unlock → TSL delta → placeholder covered → parked reads drain.

> **Note:** this branch is based on `feature/q67-placeholder` (PRs #91/#94) and **also merges `feature/q67-tsl-delta`** (PR #92), since recruitment needs both the placeholder/demand channel and `Director.apply_tsl_delta/3`.

### Recruitment flow (Distributor)

- `{:coverage_demand, tag}` now recruits instead of just logging: resolve the shard's range from the layout snapshot, pick a materializer-capable node (same convention as `MaterializerBootstrapPhase`), create a worker via that node's Foreman, `lock_for_recovery` at the current epoch, `unlock_after_recovery` with the recovery `durable_version` and a TSL filtered to the shard's logs, then `Director.apply_tsl_delta/3` and `notify_covered` to the placeholder.
- Recruitment runs off the message loop in a task; outcomes are cast back so state mutation stays serialized. In-flight demands are deduped.
- On failure: `notify_coverage_failed` (parked reads shed `:unavailable`) plus a per-tag backoff window — re-demands inside the window are ignored, and the backoff expires so later demand retries.
- A TSL delta rejected with `:newer_epoch_exists` stops the distributor (`:normal`) — it has been superseded.
- The director now hands the distributor its full TSL snapshot, the recovery `durable_version` (completed recovery attempt is retained on director state), and `node_capabilities`.

### Audit items from the ticket comments

- Director retries distributor start via `Process.send_after` when the start genuinely fails.
- Director skips re-recruitment when the distributor exits with reason `:normal` (deliberate cede to a newer epoch).
- Distributor logs older-epoch `notify_epoch_change` notifications as suspicious.
- **Not done:** structured `{:error, {:uncovered_shard, tag, range}}` from StorageRacing — `lookup_key!` has no tag at hand and the failure-map shape ripples through PointReads/RangeReads and their callers; with placeholders filling every slot the `{_range, []}` branch no longer occurs, so the loud raise stays as the backstop. Left for a follow-up.

### Telemetry

`[:bedrock, :distributor, :recruitment, :started | :succeeded | :failed]` with duration, tag, epoch, and node.

### Tests

- Unit (`recruitment_test.exs`): demand → placement/foreman/lock/unlock/TSL-delta with the bootstrap phase's injectable-fn seams; in-flight dedupe; failure → shed + backoff → retry after expiry; epoch supersession; suspicious epoch-change logging.
- Integration (`recruitment_integration_test.exs`): a read through the full PointReads → StorageRacing → LayoutIndex path against a placeholder-covered shard triggers demand, recruitment runs with ONLY the foreman worker-creation call stubbed (real epoch lock/unlock protocol, real TSL delta call), the SAME parked read returns the real value, and a subsequent read goes direct.
- Real materializer: one test recruits a REAL olivine materializer (tmp dir, health-report handshake, real `lock_for_recovery`/`unlock_after_recovery`) and serves a read through it — feasible because unlock with an empty log map starts the puller without live log servers.

Full suite: 2577 tests + 158 properties, 0 failures; credo --strict and dialyzer clean.

### DIFF
--- a/lib/bedrock/cluster/link/discovery.ex
+++ b/lib/bedrock/cluster/link/discovery.ex
@@ -161,12 +161,22 @@ defmodule Bedrock.Cluster.Link.Discovery do
     t
     |> Map.put(:known_coordinator, coordinator_ref)
     |> monitor_known_coordinator()
+    |> subscribe_to_tsl_updates()
     |> register_node_capabilities()
   end
 
   @spec monitor_known_coordinator(State.t()) :: State.t()
   defp monitor_known_coordinator(t) do
     Process.monitor(t.known_coordinator)
+    t
+  end
+
+  @spec subscribe_to_tsl_updates(State.t()) :: State.t()
+  defp subscribe_to_tsl_updates(t) do
+    # Subscribe for {:tsl_updated, tsl} broadcasts regardless of capabilities.
+    # Nodes registering resources are subscribed implicitly by the coordinator,
+    # but capability-less (pure client) links would otherwise never subscribe.
+    Coordinator.subscribe_to_tsl_updates(t.known_coordinator, self())
     t
   end
 
--- a/lib/bedrock/control_plane/coordinator.ex
+++ b/lib/bedrock/control_plane/coordinator.ex
@@ -45,7 +45,7 @@ defmodule Bedrock.ControlPlane.Coordinator do
   alias Bedrock.ControlPlane.Config
   alias Bedrock.ControlPlane.Config.TransactionSystemLayout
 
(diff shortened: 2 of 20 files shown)


jallum:
Superseded by #97, which collapses the Phase A stack (#91 → #92 → #94 → #95) into one linear sequence on current develop, carrying all audit corrections. Branch retained for history.

PR #97, opened 2026-07-15
[INTEGRATION] bedrock-q67 Phase A: shards heal their own coverage — placeholder reads, on-demand materializers, death healing, warm re-adoption, idle spin-down

jallum:
> ## ⚠️ Long-lived integration branch — NOT for immediate merge
> This is the collection point for the bedrock-q67 Phase A effort. Sub-PRs target **this branch** (`feature/q67-phase-a`), land only after a gating audit pass, and the whole stack merges to `develop` once the wider effort is ready. CI accepts PRs targeting this branch.

## Why

In Bedrock, a **materializer** is an in-memory, on-disk view of one shard of the keyspace — it replays the log and answers reads. Crucially, it is *rebuildable*: object storage holds the durable data, so a materializer is a cache of truth, not the truth itself. But until this branch, the system behaved as if every shard needed a live materializer forever. If a shard had none, the client-side routing index silently *dropped* the shard (range reads could skip data without noticing), recovery refused to finish if even one materializer couldn't start, and when a materializer died, nothing ever replaced it — a permanent hole.

FoundationDB solves placement with a dedicated **Data Distributor**, off the read/write path, and keeps shard→server locations as ordinary transactional data. We're building the same shape — with one superpower FDB structurally can't have: because our durability lives in object storage, a shard is allowed to have **zero** materializers and grow one on demand.

## What's on this branch

**Distributor** (`lib/bedrock/control_plane/distributor/`) — a cluster singleton, the janitor of coverage. The director recruits it *after* recovery completes; it never sits on the read or write path, so its death only pauses healing — reads and writes to warm shards continue untouched. Every shard-map change it makes is epoch-guarded: a stale distributor from a previous epoch self-terminates instead of corrupting routing.

**Placeholder materializer** — one lightweight process whose pid fills every routing slot that has no real server. It speaks the exact materializer read API, so clients can't tell the difference. A read arriving for a cold shard gets a polite "hold on, let me get someone": the read is *parked* for up to ~2s (bounded by the client's own timeout) while the placeholder signals demand. If the placeholder already knows the real server (a client with a stale routing table), it simply forwards. Budget exceeded → `:unavailable`, which clients already retry.

**On-demand recruitment** — the first read to a cold shard spins up a real materializer end to end: parked read → coverage demand → distributor picks a capable node → that node's foreman creates a worker → epoch lock/unlock handshake (the worker resumes from its own durable position) → routing update broadcast → the parked read is answered by the new server. The reader never saw an error, just a cold-start pause.

**Recovery relaxation** — recovery now insists only on the *system* shard (tag 0), which recovery itself needs. Data shards may come up covered by the placeholder and materialize on first touch. Nice side effect: every cold cluster boot now exercises the on-demand machinery for free.

**Death healing** — the distributor monitors every data-shard materializer. On death: placeholder swapped into the slot (reads park instead of hitting a corpse), then a new materializer is eagerly recruited under per-tag backoff. Coverage never has a hard hole; the worst case is a retry-then-park.

**Warm re-adoption** — after an epoch change (say the director restarted), the previous epoch's materializers are usually still running and holding state. Instead of rebuilding them, the distributor asks each known candidate "who are you?" (a `:shard_id` fact), epoch-locks the match, and takes it back. Example from the integration test: a warm-recovered cluster serves a read through a re-adopted worker with **zero** worker creations. Failure just means staying on the placeholder.

**Idle spin-down** (opt-in, default off) — a materializer with a configured `idle_timeout` that serves no reads for that long snapshots itself to object storage, removes its worker entry, and exits — freeing its disk and memory. Timeline: last read at 12:00:00, timeout 10min → 12:10:00 snapshot uploaded, worker gone, slot back on the placeholder → 12:47 someone reads the shard → parked ~1s while a fresh materializer cold-starts from the snapshot → answered. Only client reads count as activity; log-pulling keeps a shard *fresh*, not *hot*. The distributor recognizes `{:shutdown, :idle}` and deliberately does **not** eagerly re-recruit.

## End to end

**A read to an uncovered shard:** client routing table points the shard's slot at the placeholder → placeholder parks the read and casts `coverage_demand` → distributor picks a node, foreman starts a worker, epoch handshake, routing delta applied and broadcast → placeholder drains its queue against the new pid → client gets its value, ~1–2s late, no error.

**A materializer dies:** distributor's monitor fires → placeholder notified (stops forwarding to the corpse, parks instead) → placeholder swapped into the routing slot → re-recruitment starts under backoff → in-flight clients racing the dead pid get `:unavailable`, retry, land on the placeholder, and are answered when the recruit comes up.

## Audit discipline

Every constituent PR (#92/94/95 pre-integration, then #98–#103) was double-audited before merging here. That process found and fixed **fourteen real bugs**, including: stale pre-recovery TSLs handed to new subscribers, a coordinator epoch regression from a deposed director, a placeholder timer leak, forwarding parked reads to a dead pid, a stuck demand flag that permanently suppressed re-demand, orphaned workers on failed recruitment, a placeholder-restart race that regressed a real pid into a permanent coverage hole, a double-recruitment that orphaned a live recruit, an unbounded re-adoption lock that could pin a shard's fallback forever, and a git auto-merge that silently duplicated a handler clause.

## Deliberately interim

The **TSL delta channel** (`Director.apply_tsl_delta` + `tsl_updated` broadcast) is scaffolding. Phase B (bedrock-q67.9) moves the shard map out of the transaction system layout into the `\xFF` system keyspace — FDB's `keyServers/` shape — and *deletes* that channel and its hardening.

## Incoming

- **bedrock-q67.14** — plumb `idle_timeout` from cluster config to recruitment `worker_params` (plus the manifest-asymmetry policy decisions from the #101/#102 audits)
- **bedrock-q67.12** — system-shard (tag 0) death healing
- **bedrock-q67.11** — orphan-worker reconciliation
- **Phase B** — shard map to system keys; materializer pull source Log → Demux

Epic: `bw show bedrock-q67`.


jallum:
Closing as a merge vehicle — superseded, not abandoned.

The metadata-pipeline half of this branch (bedrock-q67.15/.16/.17/.18, bedrock-ri40) has been re-extracted onto develop as bedrock-q67.20: integration PR #152 with stacked layers #144/#145/#146/#148/#151, each re-cut against the qzr-era develop and independently re-audited.

The coverage half — the Distributor, placeholder materializer, on-demand recruitment, death healing, recovery relaxation, idle spin-down, warm re-adoption — remains valuable and is NOT superseded. It will be ported to develop under a fresh PR once the shard map moves to system-key authority (bedrock-q67.9), so the Distributor publishes coverage via transactional system-key writes from day one instead of this branch's interim TSL-delta channel (which q67.9 deletes).

The branch stays until that port lands; it is the audited reference implementation.

PR #98, opened 2026-07-15
Relax recovery to placeholder coverage for data shards (bedrock-q67.6)

jallum:
Part of the q67 Phase A effort; targets the integration branch `feature/q67-phase-a`. Closes bedrock-q67.6.

## What

Two-part recovery relaxation (per the ticket's design note — the placeholder pid is unknown during recovery, since the Distributor starts after it):

1. **Bootstrap-phase relaxation** — `MaterializerBootstrapPhase` stalls only when the SYSTEM shard's (tag 0) materializer cannot be created. A failed data shard logs a warning, emits `[:bedrock, :recovery, :bootstrap_shard_skipped]` (tag, reason), and leaves its `shard_materializers` slot ABSENT; recovery completes.
2. **Distributor coverage sweep at startup** — the distributor's first act (`handle_continue(:coverage_sweep, ...)`) diffs its shard-layout snapshot against the `shard_materializers` snapshot in the TSL the director already hands it via `child_spec` (no director change needed; tag 0 counts as covered by `metadata_materializer`, mirroring `LayoutIndex` routing), and publishes the placeholder pid into every uncovered slot as **one** batched `Director.apply_tsl_delta`. Recruitment stays demand-driven — the sweep never recruits.

## Design decisions

- **Sweep runs in a Task off the message loop**: the director blocks in `check_epoch` immediately after starting the distributor, so an in-line `apply_tsl_delta` call from init/continue would deadlock. The outcome is cast back as `{:sweep_complete, tags, result}`.
- **Epoch safety**: the sweep delta goes through the existing epoch-guarded `apply_tsl_delta`; on `{:error, :newer_epoch_exists}` the distributor stops `:normal` (check_epoch semantics). Other sweep failures degrade with a warning — the q67.1 LayoutIndex loud-failure backstop still covers those slots.
- **Placeholder restart** (stale-pid gap, previously unhandled): swept tags are tracked in `state.placeholder_tags` (added optimistically at publish time; removed when recruitment lands a real materializer in the slot). On restart the new placeholder pid is republished into exactly those slots.
- **Telemetry**: `[:bedrock, :recovery, :bootstrap_shard_skipped]` and `[:bedrock, :distributor, :coverage_sweep]` (`%{uncovered: count}`).

## Tests (TDD)

- Bootstrap: tag-0 failure still stalls; data-shard failure completes recovery with the slot absent + telemetry (same injectable seams as existing phase tests).
- Sweep: one batched delta for uncovered tags only, covered tags untouched; no delta when fully covered; stale-epoch rejection stops the distributor; placeholder restart republishes the new pid.
- Integration (crown jewel, `recruitment_integration_test.exs`): recovery-like state with an uncovered data shard → sweep publishes the placeholder → client `LayoutIndex` built from the sweep delta → `PointReads` read parks at the placeholder → demand → stub-foreman recruitment → the SAME read is served.
- Also fixes a latent test bug: recruitment_test's stale-epoch `CaptureDirector` got its init args as a single list, crashed on the first delta, and passed via director-DOWN rather than the rejection path. Hardens a few 100ms `assert_receive` budgets that flake under full-suite load (same family as a5dff723).

## Verification

- New/changed tests green at seeds 1 and 987654; `test/bedrock/control_plane` + `test/bedrock/cluster` green; full suite green (2587 tests) on an uncontended machine.
- `mix format`, `credo --strict`, `dialyzer`: clean (pre-commit enforced).


jallum:
Accepted and merged into the integration branch (gating audit passed, fix 3f606b30 included).

PR #108, opened 2026-07-16
fix: record missing point reads as conflicts

ccarvalho-eng:
## What changed

- Record regular point reads that miss storage as read conflicts by normalizing `:not_found` into the same keyed nil result path used by successful point reads.
- Convert numeric Olivine materializer shard IDs to object-storage shard tags before snapshot lookup, while keeping existing binary shard-tag startup behavior intact.

Related: bedrock-kv/job_queue#8
Found while validating dark-trench/squidie#392 for dark-trench/squidie#384 with the Squidie Bedrock sample app multi-node workflow QA.

## Why

Squidie multi-node Job Queue QA exposed that two concurrent transactions could both read an absent key, both write that key, and both commit on their first attempt. The missing read was not entering the transaction read conflict set, so the resolver had no overlap to abort and retry one writer.

The latest Bedrock branch also blocked that QA path during materializer cold start because recovery passes numeric shard IDs while snapshot object-storage paths expect binary shard tags.

## Validation

- Bedrock targeted regressions passed for missing point-read conflicts and numeric shard snapshot startup.
- Job Queue full test suite passed against the local Bedrock fix.
- Squidie Bedrock sample QA passed for the concurrent transaction conflict probe and multi-node workflow delivery against local latest Bedrock and Job Queue checkouts.


jallum:
Replace with a repair in a different location in #109 (includes same repair for key-selectors as well.)

PR #120, opened 2026-08-15
Forward drained Shale pushes to Demux (bedrock-qzr.17)

jallum:
## Summary

- return every successfully WAL-appended transaction from Shale Pushing in predecessor-chain order, including the durable prefix and valid state on partial failure
- forward each original encoded binary through the live Demux without slicing or copying in Shale
- advance KCV independently with monotonic max for parked future pushes, without changing transaction high-water, while preserving the WAL-fsync ACK boundary
- cover real Shale/ShardServer delivery, partial errors, KCV-only cut release, shuffled concurrency, and byte-identical replica chunk proposals
- document predecessor-chain ordering and the separation between transaction progress and KCV

## Verification

- `mix test` — 13 doctests, 158 properties, 2490 tests, 0 failures
- `mix quality` — formatting, strict Credo, and Dialyzer all pass

## Stack

Stack #119; targets PR #118 (`worktree-qzr16-shard-ownership`), which targets PR #117.

Ticket: `bedrock-qzr.17`

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a></sub>


lib/bedrock/data_plane/log/shale/pushing.ex
```diff
@@ -82,22 +91,27 @@ defmodule Bedrock.DataPlane.Log.Shale.Pushing do
...
-            active_segment: new_segment,
-            segments: if(t.active_segment, do: [t.active_segment | t.segments], else: t.segments)
-        },
-        encoded_transaction
-      )
+    case Segment.allocate_from_recycler(t.segment_recycler, t.path, version) do
+      {:ok, new_segment} ->
+        case Writer.open(new_segment.path) do
+          {:ok, new_writer} ->
+            write_encoded_transaction(
```
jallum, line 98: this could be restructured as a `with` block

jallum: Addressed. The allocate/open nesting this points at was flattened into a `with` when the backpressure admission landed (bedrock-qzr.23, now in #126's history), and the remaining nested `case`s that survived into the staged-rollover path are now a `with` too (db13ea3f on #131): `Writer.open` and the first append chain in one `with`, with the staged writer closed at its point of failure so the shared `else` only recycles the staged segment.

PR #223, opened 2026-08-27
Remove wait_for_healthy, the only reader of a value nothing reads

jallum:
Closes bedrock-gu0.4 and bedrock-gu0.1, under epic bedrock-gu0.

> **Housekeeping:** this PR was conflicting and had come to carry two unrelated changes, so it has been split. The conflicts were an artifact of stacking: #222 squash-merged to `develop`, leaving this branch holding a byte-identical but historically-unrelated copy of the same commit, which later edits made unmergeable. This branch is now rebased onto `develop` and contains **only** the wait_for_healthy removal. The guides terminology rename (reviewed as #224, whose merge landed on this branch instead of `develop`) now lives in #227, stacked on this one.

### Why

`Foreman.wait_for_healthy/2` has no callers — verified across lib, test, guides, benchmarks, livebooks and the sibling `bedrock_ex` checkout. The only occurrences were its own definition, its own tests, and three guide lines.

It was also the **only consumer of the foreman's health value**. `State.health` was read in exactly two places, both serving it.

That is why this family of bugs needed four tickets to find. An order-dependent fold (#214), a verdict never recomputed at boot (#218), health recomputed without waking waiters (#219), and `{:ok, pid}` outliving its process (#222) were all defects in a value whose only reader was an API nobody called. A blocking call defaulting to `:infinity`, with no callers, documenting a guarantee the code never honoured, is worse than nothing — it invites someone to depend on it.

### What changed

The function goes, and with it the `waiting_for_healthy` list and helpers, and `settle_health/1`, which collapses back to `recompute_health/1`. That undoes half of #219 by design: the pairing existed to wake waiters and there are none left. What #219 was actually worth — recomputing on *every* path that alters the worker set, including `do_new_worker/4`, which recomputed nothing — stays.

`bedrock-gu0.1` (unpruned waiter list) closes by removing the list.

### How it works now

The health verdict is still recomputed on every path that alters the worker set — deleting the blocking API deletes a false promise, not observability. The verdict simply has no reader yet; giving it one (logging, cluster monitoring) is bedrock-1a6, now the epic's only remaining observability work.

### A coverage hole closed on the way

The audit mutation-tested all seven recompute sites. Six had a test that failed when the recompute was removed; `do_worker_health/3` had **none** — its recompute could be deleted with the entire suite green. That is the one path a worker actually drives in production (Olivine's `report_health`). Now covered.

### Guide corrections

Everything the text asserted that wasn't true, not just the deleted function:

| claim | reality |
|---|---|
| `wait_for_healthy/2` gates recovery | recovery validates by locking a worker into the epoch and reading back recovery info; foreman health gates nothing |
| health transitions are logged and available to monitoring | they are not — that's bedrock-1a6 |
| `Foreman.storage_workers/1` | doesn't exist; `materializer_workers/1` does |
| a `:storage` worker kind | the kinds are `:log` and `:materializer` |
| `report_health(…, :ok)` | `Worker.health()` is `{:ok, pid} \| :stopped \| {:error, …}`; a bare `:ok` folds to `:unknown` |

The audit also caught **two overstatements in my own new text** — that a dead worker stops being advertised (true only of the roll call; the Coordinator's standing registration is withdrawn on deliberate removal, not death) and that a replacement is adopted (only if it registers inside the recheck window). Both now state the limit.

### Deliberately not done

The guides still call the materializer "storage" in prose — that rename is exactly what #227 does, stacked on this PR.

### Verification

Re-verified after the rebase onto `develop`: 2848 tests, 0 failures (one PlaceholderTest timing flake passed on immediate rerun); `mix compile --warnings-as-errors` and `mix docs` clean.


jallum (commented):
fine

PR #227, opened 2026-09-03
Guides: the component is a Materializer, and Storage means something else

jallum:
Closes bedrock-5tq.4. **Stacked on #223**; GitHub retargets to `develop` when that merges.

> **Provenance:** this is the second half of splitting the old #223, which had come to carry two unrelated changes. This content was reviewed as #224 — but #224's base was #223's *branch*, so its merge landed there instead of standing alone against `develop`. Re-landed here unchanged (verified by tree comparison), on top of the rebased #223.

### Why

The component is `Bedrock.DataPlane.Materializer`, the worker kind is `:materializer`, the capability is `:materializer`. The guides still called it Storage everywhere. bedrock-5tq.4 asked for the canonical term to be *decided* and applied — this decides for **Materializer**, because the name in the code is the one a reader can grep for.

And this isn't a terminology preference: **"Storage" already meant two different things in these guides.** 33 mentions were *object storage* — the S3/filesystem durability substrate holding chunks and snapshots — and the rest were the materializer under its old name. One word for the thing that serves reads and the thing its history is read from.

The glossary had reached the point of defining them in a loop:

> **Materializer** — "the codebase's name for a Storage server"
> **Storage** — "called a Materializer in the codebase"

Neither said what it *was*.

### What changed

`storage.md` → `materializer.md`, its inbound links and the `mix.exs` docs entry follow, `#storage` anchors resolve to `#materializer`. One canonical glossary entry, with the FoundationDB mapping a reader arriving from FDB needs. A new **Object Storage** entry — required, because the distinction being drawn needs both halves defined, and the "not to be confused with" links otherwise pointed at an anchor that didn't exist.

**Deliberately untouched:** *object storage*, *persistent storage locations*, *storage engines*, *key-value storage interfaces*, and the `:storage_read_failed` atom — which genuinely names an object-storage read. Renaming those would be the same conflation in the other direction.

### Three false claims fixed, two of them mine

The audit caught that I'd shipped exactly the kind of wrong doc claim this pass exists to remove:

- I wrote *"the code has only ever called it a materializer."* **False** — `f1a1fa94` renamed it, and `Bedrock.Internal.TransactionBuilder.StorageRacing` and `mix bedrock.dump_storage` still carry the old name and appear in the published docs. The glossary now says so, since a reader grepping for "storage" will land there.
- I wrote that a materializer *"catches up by reading [chunks]"*. **False** — a ShardServer serves its history from chunks; the materializer never replays the WAL and never reaches object storage itself. Its only contact with a log is discovery.
- `cluster.md` documented a `:storage` capability. That atom was real until `f1a1fa94` and is now rejected by `module_for_capability/1`, so the example described a node that raises at startup.

The audit also found the rename left half-done in `transactions.md` — 21 component-sense "Storage" including a mermaid diagram with `participant Storage`, which, next to a glossary now saying "Storage is the former name", read as errors rather than old terminology. Finished there and in `link.md`, `foreman.md`, `olivine.md`.

### One thing filed rather than papered over

That same config example declared `:resolution`, rejected for a different reason: resolvers are placed by the Coordinator on coordination nodes, not advertised by a node. `Cluster.capability/0` still permits the atom the supervisor rejects — **bedrock-0s9**.

### Verification

`mix docs` warning-free before and after — which is what caught two relative `[Materializer](storage.md)` links a path-prefixed grep missed. Re-verified after the split rebase: 2848 tests, 0 failures; `--warnings-as-errors` clean; `mix docs` warning-free; no `storage.md` reference remains in the guides.


jallum (commented):
fine.

PR #228, opened 2026-09-05
Give relay-chain receive assertions a window scheduling can't miss

jallum:
Closes bedrock-n9w.

### Why

`PlaceholderTest "at most one coverage demand per tag until covered"` failed on `main` (run 33218502281, OTP 27.3, the 0.7.0 release merge) and again in a local full-suite run two days later — both times as *"Assertion failed, no matching message after 100ms / The process mailbox is empty."*

The mailbox being **empty** is the tell. The demand was not lost, duplicated, or mis-addressed — it was late. Every assertion in this class waits on a chain the scheduler must run before the message can arrive: a `Task` spawn, then a `GenServer.call` into the placeholder, then the placeholder's cast back to the test (or a telemetry handler's `send`, or a spawned stub relaying). Three process hops against ExUnit's default 100ms window, on a CI box running the whole suite in parallel.

### The diagnosis is proven, not presumed

A spike made the race deterministic: suspend the placeholder with `:sys.suspend/1`, issue the gets, resume it 150ms later from a helper process. That reproduces the CI failure **byte-for-byte** — same message, same empty mailbox — and the identical scenario passes with a generous window. If late delivery had *not* reproduced the failure, the diagnosis would have been wrong and something real would have been hiding under the timeout. (The spike is not landed: it tests the assertion budget, not the product.)

Two prior data points calibrate the window. This file had already been bitten once — the coverage-failure test carried `, 1_000` with the comment *"the default 100ms flakes under full-suite load"* — one site hardened, fourteen siblings left bare, and a bare sibling is what failed `main`. And bedrock-uvk records a **1.5s** window being missed under full-suite load elsewhere. Hence 5s: an `assert_receive` window is free when the message is prompt; it only delays the report of an already-failing test.

### What changed

All 36 relay-chain `assert_receive` sites across `placeholder_test.exs` and `server_test.exs` now carry `, 5_000` (the repo's idiom is inline numeric windows — there is deliberately no global `assert_receive_timeout` override, matching how bedrock-yp2/bedrock-c4u fixed each mechanism at its sites). The previously-hardened `1_000` site is normalized into the class, and one comment at the first site records the mechanism.

### Deliberately untouched — different mechanisms, not oversights

- **`assert_received` sites** — the message was sent synchronously before the assertion; there is no wait to widen.
- **The `:noproc` `:DOWN` assert** — that monitor fires without any process running, so there is no relay to schedule; the races it *can* have are reason races, bedrock-yp2 territory (guards, not windows). The *other* `:DOWN` assert — whose monitored recruit task must be scheduled, run, and crash before the DOWN exists — is the relay class and IS widened.
- **`refute_receive` absence windows (50ms)** — dedupe is state-based in `signal_demand/2` (a `MapSet` check, not timing), so 50ms proves what it needs to; widening only slows the suite.
- **Timer-tick asserts** (`:poll_lock`, `:reverify_assignment`, `:retire_member`) — `Process.send_after` delivers from the timer wheel straight into the test process's own mailbox, no relay to schedule, with 20-40x margin over the tick intervals.

### Verification and review

Distributor tests repeated 30× green (`--repeat-until-failure 30`); full suite 2835 tests, 0 failures; `mix format --check-formatted`, `credo --strict`, and dialyzer all clean.

Cold-reviewed by a context-free agent before this PR was opened. Verdict: approve — every widened site verified as a pure existence assertion, the dedupe `refute_receive` interplay and the `hold_ms` deadline tests confirmed unweakened. The review also caught two exclusions of mine that did not survive scrutiny — the `RuntimeError` `:DOWN` (relay-class after all: the monitored task must run and crash) and the grandfathered 500ms recruitment window (undercut by the same bedrock-uvk datum that justified 5s) — both folded into the class above.

### DIFF
--- a/test/bedrock/control_plane/distributor/placeholder_test.exs
+++ b/test/bedrock/control_plane/distributor/placeholder_test.exs
@@ -68,13 +68,19 @@ defmodule Bedrock.ControlPlane.Distributor.PlaceholderTest do
 
       task = async_get(placeholder, "apple", timeout: 5_000)
 
-      assert_receive {:"$gen_cast", {:coverage_demand, 1}}
-      assert_receive {:telemetry, [:bedrock, :distributor, :placeholder, :parked], %{count: 1}, %{tag: 1}}
+      # Receive windows in this file are generous (5s): every asserted
+      # message needs at least one other process — a Task, the
+      # placeholder, or a telemetry relay — scheduled before it can
+      # arrive, and the default 100ms window flakes under full-suite
+      # load (bedrock-n9w). assert_receive returns the moment the
+      # message arrives, so the width is free.
+      assert_receive {:"$gen_cast", {:coverage_demand, 1}}, 5_000
+      assert_receive {:telemetry, [:bedrock, :distributor, :placeholder, :parked], %{count: 1}, %{tag: 1}}, 5_000
 
       :ok = Placeholder.notify_covered(placeholder, 1, stub)
 
       assert {:ok, "red"} = Task.await(task, 5_000)
-      assert_receive {:telemetry, [:bedrock, :distributor, :placeholder, :drained], %{count: 1}, %{tag: 1}}
+      assert_receive {:telemetry, [:bedrock, :distributor, :placeholder, :drained], %{count: 1}, %{tag: 1}}, 5_000
     end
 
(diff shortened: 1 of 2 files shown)


jallum (commented):
accepted

PR #230, opened 2026-09-05
Adopt bedrock_raft 0.10 and flatbuffer 0.6; extend DiskRaftLog to the new Log contract

jallum:
Re-opens the dependency stack against `develop`. #225 was closed unmerged and #226 was then merged *into* this branch, so the whole stack has been sitting off `develop` since 2026-09-03 — `develop` still pins `{:bedrock_raft, "~> 0.9"}` / `{:flatbuffer, "~> 0.5"}`. This PR lands it. Closes bedrock-54w and bedrock-fu8.

Two commits: a mix.lock refresh, and the raft/flatbuffer major bumps with the code they require.

## Why

`bedrock_raft` 0.10.0 extends the `Bedrock.Raft.Log` protocol and changes the shape of the ack telemetry event, so the coordinator's `DiskRaftLog` no longer satisfies the contract it implements. Three of the additions exist for real reasons we want:

- **Votes must survive a crash.** Pre-0.10 the log persisted only `current_term`; a node that voted, crashed, and came back could vote again in the same term. 0.10 makes `voted_for/1` + `save_election_state/3` part of the protocol so term and vote are written together.
- **Replication is batched.** `transactions_from/4` lets the leader pull one bounded batch per AppendEntries instead of everything from the follower's cursor to the tail.
- **Purges must be honest.** The leader now repositions its send cursor from a follower's hint via `has_transaction_id?/2`, so the log can't leave purged records physically present.

Separately, `flatbuffer` 0.6 moves to compile-time, schema-specialized codegen. 0.6.1 is what makes it adoptable cleanly (see below), and with it every dep resolves from hex — the interim git-tag dep on `bedrock_raft` is gone (`~> 0.10`), so nothing blocks `mix hex.publish`.

| dep | from | to |
|---|---|---|
| bedrock_raft | 0.9.7 | 0.10.0 |
| flatbuffer | 0.5.0 | 0.6.1 |
| iodata | 0.7.3 | 0.9.0 |
| telemetry | 1.3.0 | 1.4.2 (unblocked by raft relaxing its pin to `~> 1.3`) |

Plus a routine lock refresh of dev/test tooling with no `mix.exs` requirement changes: req 0.7.4, mox 1.3.0, stream_data 1.4.0, styler 1.12.2, benchee 1.5.1, ex_doc 0.40.3, mix_test_watch 1.4.0.

## What changed

Two production modules do the work, plus a small cleanup in a third:

- `DiskRaftLog` (+ its `defimpl`) — new `voted_for/1`, `save_election_state/3`, `previous_transaction_id/2`, `transactions_from/4`; reworked `purge_transactions_after/2` and `save_current_term/2`.
- `RaftTelemetry` — ack tracing clauses moved to the 0.10 metadata keys.
- `ClusterBootstrap` — drops a dialyzer suppression and a now-redundant `@external_resource`.

TDD'd: the protocol contract was pinned in `disk_raft_log_protocol_test.exs` first (26 red → green), and both ack shapes are pinned in `raft_telemetry_test.exs`.

Nothing else needed to move. The wire event change (`:append_entries_ack` is now a 5-tuple) passes through `Raft.handle_event/3` opaquely, and `Bedrock.Raft.Interface` is unchanged, so `RaftAdapter` is untouched.

## How it works

**Election state.** Term and vote live in a single `{:election_state, {term, voted_for}}` DETS record, so they're persisted in one write, as the protocol's "atomically persist" requirement demands. Reads fall back to a pre-0.10 `{:current_term, term}` record (as `{term, nil}`) so existing coordinator logs open without migration. `save_election_state/3` enforces the safety rules: advancing the term may set any vote; at the durable term a vote may be set once or repeated verbatim but never changed or cleared (`{:error, :already_voted}`); below it, `{:error, :stale_term}`. `save_current_term/2` now only acts when strictly advancing, and clears the old term's vote when it does — matching the in-memory logs bundled with raft.

**Bounded reads.** `transactions_from/4` and the underlying `walk_chain_inclusive/4` take a count that decrements per chain hop, with `:infinity` preserving the old unbounded behavior; `transactions_from/3` delegates with `:infinity`, so existing callers are unchanged.

**Purge.** A purge below the commit point is rejected outright (`{:error, :would_delete_committed_transactions}`) — Raft's commit index must never decrease — and a purge at or past the tail is a no-op. Otherwise the purged transactions and their chain links are *deleted*, not just unlinked, because under 0.10 a stale physical record would answer `has_transaction_id?/2` for an entry the chain can no longer reach and mis-route replication.

**`previous_transaction_id/2`.** A single-pass `:dets.select` with `is_integer` guards on both halves of the key, which picks out exactly the transaction records (chain links are keyed `{:chain, id}`, metadata by a bare atom). Documented deviation: the protocol asks for O(log n); DETS is unordered, so this is O(n). The coordinator log is small and the leader only consults this on a rejected ack, so it isn't a hot path.

**Telemetry.** `RaftTelemetry.log_event/4` has no catch-all clause, so a stale match would raise inside the handler and `:telemetry` would silently detach it — hence the ack clauses move to `success`, `request_transaction_id`, and `follower_newest_transaction_id`, with tests pinning both ack shapes.

**flatbuffer 0.6.1, not 0.6.0.** The public `use Flatbuffer` API is unchanged, but 0.6.0's generated code carried clauses for every scalar type, which produced 17 unreachable-code dialyzer warnings in `ClusterBootstrap`; a module-wide suppression papered over that for one commit. 0.6.1 prunes codegen to the types reachable from the schema root (upstream fb-3lk), so dialyzer is clean and the suppression is gone again. 0.6 also declares the `.fbs` as `@external_resource` itself, so our manual declaration and its stale comment are removed — verified that editing schema content still recompiles the module.

## Verification

`mix compile --warnings-as-errors`, `mix format --check-formatted`, `mix credo --strict`, dialyzer (0 errors), full suite: **2878 tests, 158 properties, 0 failures** — as run on the branch. `develop` has since advanced by #223 and #228, neither of which touches this area; CI on the merge result is the check that matters.


jallum (commented):
lfg

PR #240, opened 2026-09-06
Drop deleted ChunkWriter from LocalFilesystem moduledoc

jallum:
Closes bedrock-jzh.

Doc-only. The `put_if_not_exists/4` caller list in `LocalFilesystem`'s moduledoc still named `ChunkWriter`, which #220 (bedrock-d3k) deleted. Verified the surviving callers by grep: `Demux.ShardServer`, `Snapshot`, `ClusterBootstrap.Discovery`, `PersistencePhase` — all still listed accurately.

Found while auditing the tracker for tickets whose work had merged without being closed (15 closed, 23 stale worktrees removed).


jallum (commented):
lfg

PR #254, opened 2026-09-06
237: Rebuild bootstrap coverage after materializer retirement

### LINKED ISSUE #237: Tag-0 retirement can leave CoreState pointing only to unavailable materializers
Beadwork: `bedrock-q67.21.20`

Tag 0 has two membership authorities: recovery writes the bootstrap record in object storage, while the distributor can change the committed `materializers/0/*` set. They can diverge so the next recovery stalls despite a healthy system-shard materializer.

Source-level finding at `8f962e09`; not yet reproduced as a full-cluster failure. This is the same defect already described by the linked Beadwork ticket.

## Failure sequence

1. Recovery records system materializer `w1` in CoreState and the keyspace.
2. Distributor membership changes introduce healthy `w2` for tag 0.
3. `w1` dies and the distributor retires it; keyspace membership is now `w2`, while CoreState still names `w1`.
4. The next recovery finds no available CoreState-named member and stalls with `:system_materializers_unavailable`.
5. It cannot read the newer keyspace membership because finding a system materializer is prerequisite to reading that keyspace.

See [lib/bedrock/control_plane/distributor/server.ex](https://github.com/bedrock-kv/bedrock/blob/8f962e09b24765265eeb6bd818f5b26da8d91b99/lib/bedrock/control_plane/distributor/server.ex#L611) and [lib/bedrock/control_plane/director/recovery/materializer_bootstrap_phase.ex](https://github.com/bedrock-kv/bedrock/blob/8f962e09b24765265eeb6bd818f5b26da8d91b99/lib/bedrock/control_plane/director/recovery/materializer_bootstrap_phase.ex#L372).

## Required design decision

Establish one crash-safe authority for bootstrap membership. The existing ticket explores recovery-only ownership, restricted distributor membership changes, a safe pointer publication protocol, or an independent bootstrap metadata source. Do not fix this by blindly accepting an arbitrary tag-0 claimant or assuming object-store and keyspace writes are atomic.

## Acceptance criteria

- Reproduce retirement/replacement of a CoreState-named member followed by an epoch change.
- Recovery can find an authoritative surviving bootstrap source or fails for a genuine availability reason.
- Exercise crashes at every publication boundary and stale/displaced workers.
- Preserve the distinction between durable history and materializer read coverage.

jallum:
Recovery reads the shard layout from tag 0, finding a tag-0 materializer by the names it wrote into the object-storage bootstrap record. The distributor also manages tag-0 membership between recoveries, so it can retire the exact worker that record names, and recovery then stalled with a healthy replacement serving the shard. Those names are now a cache preference; a missing or stale cache is rebuilt from durable history.

Ticket: bedrock-q67.21.20
Closes #237

## Why

The shard boundaries and the materializer membership both live in tag 0, so recovery cannot read either until a tag-0 materializer is running. On the base it took the names from the bootstrap record, kept only workers this epoch had locked, and stalled if none survived. That rule came from FoundationDB, whose recorded server set changes only during recovery; Bedrock's changes constantly, and the distributor never touches the bootstrap record.

The resulting failure is all healthy behaviour: recovery records `w1`; the distributor recruits `w2`, then `w1` dies and is retired, leaving the committed family naming only `w2`. The next recovery looks for `w1`, does not find it, and cannot consult the keyspace to learn about `w2`, because reading that keyspace is what it was trying to bootstrap. A survivor withdrawn from the family hit the mirror bug: it was adopted and republished, resurrecting a dropped assignment.

## What changed

Authority moves from the recorded names to the durable stream behind them. A tag-0 materializer is only a read cache: `ShardServer` writes shard-keyed, epoch-spanning chunks that are never deleted, and log replay copies the recovered WAL suffix.

Operator-visible: three stall reasons are deleted, `{:system_materializers_unavailable, named}`, `:no_system_materializer_found`, and `{:ambiguous_system_materializer, ids}`. Those conditions reconstruct instead of stalling, and a stall now names the stage that failed. The distributor keeps tag-0 coverage, and no cross-store write ordering is assumed.

## How it works

A named worker is still preferred, since a warm cache saves replay. Recovery locks and unlocks it at the recovery version, reads the layout and membership at that version, and keeps it only if the membership it just read still names that worker on that worker's node. Otherwise a fresh tag-0 worker is recruited. Its new-epoch `ShardServer` has no durable watermark, so its pull serves retained chunks first and the replayed suffix after, reaching the committed layout rather than an empty one.

```
record names {w1}; committed family names {w2}
lock w1 -> gone  =>  recruit w3
w3 pulls chunk@100, then replayed 200, 300
read membership at 300 -> {w2}; publish {w2, w3}
```

Only a worker this phase created and has not yet offered to persistence is removed on failure; once the recovery transaction commits, a failed bootstrap write leaves it running for membership reconciliation. A lock failure on a named worker, such as a newer-epoch fence, still stalls rather than reconstructing around it.

## Verification

A new integration test reproduces that divergence against real chunk history, real Olivine materializers and the production distributor handlers, then runs bootstrap across four cache states (dead, displaced, snapshot-seeded, legacy nameless) and five publication crash boundaries. On the base the dead cases stalled and the displaced cases re-adopted the withdrawn worker. Unit tests cover cleanup of an unpublished worker at each failing stage. CI on `cc09b0dc` is green on OTP 27.3, 28.3 and 29.0 plus the MinIO S3 job.

### DIFF
--- a/guides/quick-reads/recovery/materializer-bootstrap.md
+++ b/guides/quick-reads/recovery/materializer-bootstrap.md
@@ -1,54 +1,53 @@
 # Materializer Bootstrap
 
-**Get the shard layout back, by asking the materializer that already holds it.**
+**Read the committed shard layout from a verified cache or rebuild that view from durable history.**
 
-Recovery needs to know the cluster's shard boundaries before it can place
-resolvers or route anything. Those boundaries live in the committed keyspace
-under `\xff/system/shard_keys/`, which means recovery has to bring one
-materializer — the one serving the system shard, tag 0 — back online and read
-them out of it.
-
-This phase is deliberately *not* a recruitment phase. Materializers hold
-irreplaceable committed state, so the goal is to reuse the survivors, not to
-replace them.
+The shard boundaries live in `\xff/system/shard_keys/*`, served by tag 0.
+The materializer serving that view is disposable: recovered logs and
+object-storage chunks hold the history. Recovery must make the view readable
+at its chosen recovery version before placing resolvers or routing writes.
 
 ## Fresh Cluster
 
-With no prior logs there is nothing to recover, so the phase invents the
-starting layout: two shards, the system shard (tag 0) covering `0xFF` to the
-end of the keyspace, and the user shard (tag 1) covering everything below
-`0xFF`.
+A bootstrap record with no prior logs means there is no committed history.
+Recovery creates a system materializer and seeds two shards: tag 0 for system
+keys and tag 1 for user keys. Only this fresh path invents a layout.
 
 ## Existing Cluster
 
-1. **Resolve the system materializer by name** from the prior
-   [`CoreState`](../transaction-system-layout.md). It is looked up, never
-   invented — recovery does not guess which worker was serving tag 0.
-2. **Lock it** for this recovery.
-3. **Unlock it** with its replica set of pull sources, so it starts pulling
-   from the logs.
-4. **Wait for catchup** — up to 60 seconds, polling — until it has applied
-   through the recovery version.
-5. **Read the shard layout** from `\xff/system/shard_keys/*`.
+1. Prefer a locked tag-0 cache named by CoreState. Unlock it at the recovery
+   version and let it resume its own stream cursor.
+2. Wait until it can serve that version, then read the shard layout and
+   materializer membership at the same version. Reuse the cache only if the
+   committed family still names its worker and node.
+3. If no named cache is available or the named cache was displaced, recruit a
+   fresh tag-0 worker. Legacy records without cache hints use this path too.
+   The worker starts from a committed snapshot or zero and streams old chunks
+   followed by the recovered WAL suffix through its ShardServer.
+4. Wait for catchup and read the existing layout and membership. An empty
+   layout or failed read stalls recovery; it never becomes a fresh layout.
+5. Carry any new worker into the recovery system transaction so its read
+   coverage is published only after the historical reads succeed.
 
-The `\xff/system/materializers/` family is read alongside it: a family-named
-worker that this epoch locked, whose own shard assignment agrees, is
-re-adopted for its shard. Only a shard with no survivor gets a fresh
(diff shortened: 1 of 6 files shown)


mikehostetler:
## Current-code application verification — 2026-09-15

I adapted the implementation at `cc09b0dc00532919a94bd87b1de4b31ef89bee5b` to develop `153a2b62` plus #326 at `8d0bb980`. This is your cache-rebuild design, not a competing replacement.

The port keeps these boundaries:

- Replay a preferred named tag-0 cache through the recovery version, then validate its committed membership at that same version.
- If no named cache is valid, recruit a tag-0 cache and rebuild from durable logs/chunks; read the recovered metadata, not a default empty layout.
- Keep named-cache lock/fence failures as failures. Do not reconstruct around an epoch fence.
- Leave data-tag adoption to Distributor; remove only newly created, unpublished replacement workers after failure.

The Bedrock-only regression starts four peers with three logs / RF3, commits one binary checkpoint, and kills the node that owns both a log and the system cache. A survivor completes recovery, retains the cluster identity, reads the exact bytes, commits revision two, and rejects a stale revision. A second case loses the elected leader after followers booted without a bootstrap.

Both cases pass at seeds 0 and 7. They use layout events, process monitors, state barriers, and real transactions, without sleep timers.

**This is a combined-patch result:** #326 supplies current durable recovery input, and #334 avoids shard discovery crashing a locked Shale log. On the #326 diagnostic baseline, node-loss recovery times out; adding only the Shale guard still times out. I have not claimed that the original #254 head alone passes on current develop.

The focused recovery/log/S3 selection passes 98 tests; the full default suite passes 3,151 with 19 excluded. Jido's Bedrock files pass 33/33. One-host/shared-storage limits apply.

The adaptation and peer tests are local and uncommitted. Would you prefer the current-code adaptation on your branch, or a separate attributed follow-up after rebase? I will preserve your authorship and coordinate helper reuse with #322 before submitting test changes.

[Jido tracking #379](https://github.com/agentjido/jido/issues/379); related issue #237.

PR #278, opened 2026-09-06
bedrock-s1zr: prune a shard's snapshots past a keep-last limit

jallum:
Every snapshot an Olivine materializer wrote stayed in object storage forever: the deletion primitive existed but nothing in `lib/` called it. This change adds a keep-last-K retention policy, set by one manifest param, and prunes a shard's older snapshots right after each successful upload.

Ticket: bedrock-s1zr
Stacked on: #269 (bedrock-zi44/snapshot-upload-policy)

## Why

A materializer uploads a snapshot synchronously before an idle spin-down and in a fire-and-forget task after a compaction cutover. Both write a put-if-not-exists object under `s/{tag}/` keyed by inverted version; neither removed anything. #269 throttled how often a snapshot is written, but the objects still accumulated without bound, one per spin-down in practice (bedrock-947).

Deleting a snapshot is safe because it is derived state: anything it held can be rebuilt by replaying chunks forward from an older snapshot. Chunks stay untouched (bedrock-wxf.6.11). Retention owes one thing: a shard is never left with zero snapshots, since a cold start with no baseline replays from version zero.

## What changed

- New pure policy `SnapshotRetention` with one field, `keep_last`, read from the manifest param `snapshot_keep_last`. Only an explicit positive integer sets it.
- The knob rides the same path as `idle_timeout`: manifest params, worker startup opts, a new `State` field.
- Both upload sites prune immediately after the write returns `:ok`. No scheduled prune exists.

A worker without the param deletes nothing and issues no listing. With it set, an upload may log `Pruned snapshots past the retention limit` or warn `Snapshot retention pruning failed`. Neither changes the upload's result or the spin-down decision.

## How it works

The prune lists the shard's snapshots, newest-first because keys are inverted versions, and the policy returns the Kth newest existing version as a floor, or `:keep_all` when nothing is older. Everything strictly below the floor is deleted.

```
listing after four uploads, keep_last: 2   [4, 3, 2, 1]
floor = 2nd newest                          3
delete strictly below 3                     [4, 3]
```

The newest snapshot cannot be deleted by construction: the floor is always a version from the listing, and with `keep_last: 1` it is the newest itself. The prune runs only after the write succeeded, so the object it keeps is durable before anything is removed.

A listing that cannot complete raises rather than yielding a short list; reading it as empty could floor on a truncated history and delete a wanted snapshot while reporting success. The rescue spans the whole prune, including the second listing inside the deletion walk, because on the spin-down path an escaped raise would crash a worker that had already landed its snapshot. The upload still returns `:ok`.

Merge order matters: land after #268. Before it, a foreign object under `s/{tag}/` whose name parses as base36 but is not canonical width yields a fabricated version up to 2^64-1, which as a floor would delete every real snapshot. #268 rejects such keys at the listing.

## Verification

Integration tests drive both upload paths with `keep_last` set and assert the surviving listing, including a cold start from the lone survivor after a spin-down prune. A backend whose listing raises on consumption proves the failure is logged while the upload returns `:ok`, whether the first or only the second listing fails, and that an unconfigured worker never lists. Unit tests pin the knob and floor arithmetic. Author reported suite, format, credo and dialyzer clean; no CI checks on this stacked draft.

Follow-up: bedrock-q67.21.19 (#268), canonical key parsing, land first.
Follow-up: bedrock-947, nothing drives compaction today.
Follow-up: bedrock-wxf.6.11, chunk reclamation below replay floor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## Unreleased
 
+- **Snapshots can be pruned once they are written.** Every snapshot a
+  materializer ever wrote stayed in object storage forever; nothing in
+  `lib/` called `Snapshot.delete_older_than/2`. A new manifest param,
+  `snapshot_keep_last`, says how many of a shard's newest snapshots to
+  keep, and the prune runs after a snapshot has durably landed — so a
+  shard can never be left without a baseline, and the newest snapshot is
+  never the one deleted. A worker without the param deletes nothing and
+  does not even list. A prune that cannot complete its listing says so
+  rather than reporting nothing to delete. Shard chunks are untouched:
+  they are the history, and reclaiming them needs a replay floor
+  (bedrock-wxf.6.11).
+
 - **Snapshot uploads answer to a policy.** A materializer uploaded a
   snapshot whenever a compaction happened to finish, with no way to say how
   much of that cadence was worth paying for. Three new manifest params say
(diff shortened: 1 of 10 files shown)


jallum (commented):
lfg

PR #280, opened 2026-09-06
bedrock-q67.21.8: Production wiring for materializer idle_timeout

jallum:
Materializer idle spin-down has been dormant since it landed: nothing in production supplied the `"idle_timeout"` worker param, so every materializer ran with an infinite timeout. A cluster parameter now carries the window (default fifteen minutes, zero disables) through the bootstrap record to the distributor, so each data-shard recruit is created with it. The system shard is exempted where the distributor creates workers.

Ticket: bedrock-q67.21.8
Stacked on: #263 (bedrock-q67.21.13/recovery-stops-at-tag-0)

## Why

A materializer spins down only when its manifest carries a positive `"idle_timeout"`. The distributor's recruitment code already merged a `:worker_params` map from its context into each new worker's params, but the director built that context without the key. The plumbing existed and the source did not.

The ticket assumed the system shard was exempt by construction because recovery seats tag 0 with the shard id alone. Recovery is not the only creator of a tag-0 materializer. Tag 0 sits in the shard layout and the committed `materializers/` family like any other tag, so the distributor monitors it, heals a dead member through the same recruit path data tags use, and includes it in the startup sweep. A system materializer carrying the timeout would eventually snapshot, delete its own foreman entry, and exit; the next recovery would stall on a named system member that no longer exists.

## What changed

- `materializer_idle_timeout_ms` joins the cluster parameters, the coordinator's defaults, and the bootstrap record.
- The director's recruitment context carries `worker_params: %{"idle_timeout" => ...}` read directly from config.
- `Recruitment.worker_params/2` returns the shard id alone for tag 0 and merges the policy params for every other tag.
- Spin-down logs the worker id, shard, and idle duration alongside the telemetry event.

No operator write path for cluster parameters exists and none is added. The ticket's two observability residuals are left for their own tickets.

## How it works

The value takes the route the shard assignment already takes: cluster config to the recruitment context, to the foreman's manifest at creation, to the worker's startup opts. Only client reads count as activity. A running data-shard worker that passes the window uploads a snapshot, arranges its foreman removal, and exits `{:shutdown, :idle}`; the distributor retires the member without re-recruiting, and the next read parks behind the placeholder and revives the shard on demand.

Zero has to survive the bootstrap round trip. A flatbuffer omits a scalar equal to its schema default, and the implicit default for `uint` is zero, so a disabled cluster would read back as absent and be silently re-enabled by the coordinator's fallback. The schema declares the default as 900000 instead: zero is written, while a record from before the field existed still reads back absent and picks up the parameter default. The two defaults must stay equal, and the schema comment says so.

The timeout is frozen in the manifest at creation, so a changed parameter reaches the next recruit, not running workers. An existing cluster gets spin-down enabled on its next recovery.

## Verification

Tests pin the director's context carrying the configured value, the exact params for a data tag and for tag 0, recovery's tag-0 seating, and the zero-value bootstrap round trip. An end-to-end case sets only the cluster parameter, recruits a live Olivine worker through the real recruit path, and asserts the `{:shutdown, :idle}` exit; a sibling case with zero asserts the worker stays up. Suite, format, credo and dialyzer reported clean; no checks run on this stacked branch.


lib/bedrock/data_plane/materializer/olivine/server.ex
```diff
@@ -663,6 +665,16 @@ defmodule Bedrock.DataPlane.Materializer.Olivine.Server do
...
 
+        # Spin-down removes a shard's only copy on this node, and the
+        # next read pays for the revival: an operator reading the log
+        # after a slow first read needs to see that this happened, to
+        # which shard, and after how long — the telemetry event is for
+        # metrics, not for the postmortem.
+        Logger.info(
+          "Materializer #{inspect(t.id)} (shard #{inspect(t.shard_num)}) spinning down " <>
+            "after #{idle_ms}ms without a read"
+        )
```
jallum, line 685: We should move this log to the telemetry handler.

PR #304, opened 2026-09-06
bedrock-q67.21.14: Rename components to match their clarified purposes

jallum:
The q67.21.9 to .13 stack changed what several recovery and distributor components do, and their names kept describing the old behaviour. This is the deferred rename pass, landed last so nothing is renamed twice. No logic changes.

Ticket: bedrock-q67.21.14
Stacked on: #303 (bedrock-q67.21.15/exclusion-safety-log-record)

## Why

Names here are read as claims about the code, and five had become false. `MaterializerBootstrapPhase` claimed recovery seats materializers for every shard. Since #263 it creates or adopts exactly one, for tag 0, so that recovery can read the metadata families that live there; placing data-plane materializers is the distributor's job. `TSLValidationPhase` claimed to check a transaction system layout. It has only ever received the prior durable record, which since #200 is a `CoreState` holding `logs` and `system_materializers` and nothing else. The stall reason `:corrupted_tsl` and the `:tsl_validation_*` telemetry pointed an operator at the wrong artifact.

`materializer_refs` in the distributor snapshot and `prior_materializer_refs` on the attempt read as one callable ref per tag. Since bedrock-q67.21.9 the family is a set, `%{tag => %{worker_id => node}}`, and every reader iterates it as one. `shard_materializers` implied recovery still decides membership for many shards.

## What changed

- `MaterializerBootstrapPhase` becomes `SystemShardBootstrapPhase`; its read-failure reason is `{:prior_members_query_failed, _}` and the test hook is `:read_prior_members_fn`.
- `TSLValidationPhase` becomes `CoreStateValidationPhase`; the stall reason is `{:corrupted_core_state, _}` and the events are `[:bedrock, :recovery, :core_state_validation_success | _failed]` with metadata key `core_state`.
- `TSLTypeValidator` becomes `TypeSafetyValidator`, not the ticket's `CoreStateValidator`: its assertive half runs on the real layout in `TopologyPhase`, whose `resolvers` list is genuinely checked, so it is named for what it checks.
- `shard_materializers` becomes `seated_materializer_members`, not the ticket's `system_shard_members`: the field stays tag-keyed. `prior_materializer_refs` becomes `prior_materializer_members` rather than being deleted; #299 kept it as the routing seed and persistence diff base. The snapshot key becomes `materializer_members`.
- `SystemKeys.ClusterBootstrap` keeps its name as the wire format. Its moduledoc now states that `CoreState` is the recovery-facing projection, and that schema changes recovery must see need a matching change in `from_bootstrap/1`.

Also fixed: `TopologyPhase` docs said it transitions to `PersistencePhase`; it returns `MonitoringPhase`. The corruption log drops a `Resolvers:` line that was always `nil`.

## How it works

The data flow is unchanged; the names now match it. Recovery starts at `CoreStateValidationPhase`, which type-checks the prior record's log tag ranges (integer tags arriving as version binaries is the corruption it exists to catch) and stalls with `:corrupted_core_state` on a mismatch. `SystemShardBootstrapPhase` then brings up tag 0 and records what it seated beside what the keyspace already held:

```elixir
seated_materializer_members: %{0 => %{"wkr_sys" => "node@host"}}
prior_materializer_members:  %{0 => %{...}, 1 => %{...}}   # whole family, one version
```

Only the diff is written to `materializers/`. `TopologyPhase` merges the two to seed the proxies' routing snapshot; the persistence phase merges them at tag 0 for the bootstrap record. The distributor reads the same family into `materializer_members` and maintains it as it publishes placeholders and adopts or retires members.

## Verification

A pure rename argues its correctness through an unchanged suite: 2933 tests, 0 failures, identical to the base, with format, credo and dialyzer clean and no assertion changed beyond names. A case-insensitive grep across `lib/`, `test/` and `guides/` finds the old identifiers only in the released 0.7.0 section of `CHANGELOG.md`. No CI checks run on this stacked branch.

Follow-up: bedrock-5tq.8, guides content and release notes.


jallum (commented):
lfg

Issue #297, opened 2026-09-06
Fence Shale replay and live pushes with recovery ownership tokens


jallum:
Recovery epoch comparisons alone do not distinguish two owners in the same generation, and Shale's replay/live-push paths do not consistently carry the recovery authority that granted the lock. Delayed tasks can therefore reach a reused log without a complete ownership check.

Add an explicit generation plus granted-owner token through Shale lock, replay, unlock, live push, and delayed reply paths.

Acceptance criteria:

- Reject lower-generation and equal-generation foreign owners before any WAL or in-memory mutation.
- Thread the token through real task-based LogReplay and CommitProxy callers.
- Cover stale replay, stale live push, stale unlock, delayed pre-lock requests, duplicate same-owner requests, and delayed replies after replacement.
- Inspect actual WAL/log state and prove rejected operations make no change and report no phantom success.
- Preserve acknowledged prefixes and prevent duplicate atomic tails when the valid new owner retries.
- Audit Olivine's existing owner fence and transient component shutdown as integration consumers.
- Do not add an unfenced legacy fallback; use a compatible-binary-only coordinated upgrade policy.

Depends on the consensus generation/reservation contract and blocks final #259 integration.

Beadwork: `bedrock-50n`

Issue #296, opened 2026-09-06
Allocate recovery generations through consensus and reserve bootstrap publication


jallum:
Director replacement under an unchanged Coordinator leader currently reuses the Raft election term as the data-plane recovery epoch. A local integer increment is not durable or consensus-authoritative, and the bootstrap writer currently reads a fresh CAS token at the end of recovery, allowing a delayed older generation to overwrite a newer reservation.

Implement consensus-committed recovery generations and same-bootstrap publication reservation as one protocol unit.

Acceptance criteria:

- Allocate recovery generations through committed Raft commands after a leader barrier; never launch from historical replay or before quorum.
- Apply `:behind` and `:latest` committed callbacks monotonically from one persisted applied cursor; duplicates/regressions are no-ops.
- Reconstruct a durable generation floor across Coordinator restart, inherited commits, leadership changes, checkpointing, and truncation.
- Reserve the existing bootstrap object before Director launch while preserving prior committed metadata, using one immutable CAS token for final publication.
- Never refresh the final CAS token. If CAS result is unknown, readback only verifies the exact publication identity and payload.
- Define the CAS-success/notification-loss rule and require a successor to allocate a strictly greater generation.
- Exercise no-quorum behavior, competing leadership, actual DiskRaftLog persistence, LocalFilesystem CAS orderings, and crashes around allocation, reservation, system commit, final CAS, and notification.
- Provide an explicit coordinated quiescent schema rollout and fail closed on unsupported formats. Online mixed old/new writers are outside scope unless a separate capability barrier is implemented.

Depends on the attributed Director notification protocol. Blocks Shale ownership fencing and final #259 integration.

Beadwork: `bedrock-7i1`

Issue #295, opened 2026-09-06
Attribute Director notifications to the current recovery instance


jallum:
A delayed old Director publication can restore retired log routing after a replacement recovery. The current layout and config notification messages carry no Director identity, and the Coordinator accepts them without proving that they came from its current recovery instance.

This was reproduced from #259 with a real recovered cluster: after replacement recovery completed, replaying the exact prior Director's layout notification replaced the current log routing. Both Directors used the same epoch, but the identity flaw is independently fixable before generation allocation changes.

Acceptance criteria:

- Carry Director PID and generation attribution on layout and config notifications.
- Accept messages only from the Coordinator's current registered Director while it remains leader for that generation.
- Notifications must not assign or roll back the Coordinator epoch/generation.
- Clear and broadcast layout unavailability when retiring the current Director.
- Define an idempotent, monotonic publication identity for duplicate current-instance notifications.
- Cover stale layout, stale config, an equal-generation foreign PID, leadership loss, duplicate DOWN, and duplicate/out-of-order publication messages.
- Do not retain a permissive unattributed legacy message path.

This issue deliberately excludes recovery-generation allocation and bootstrap schema changes. It blocks the remaining #259 protocol work.

Beadwork: `bedrock-ls9`

PR #308, opened 2026-09-06
259: Cover fenced Director replacement recovery

### LINKED ISSUE #259: Director failure reuses the recovery epoch and a log crash stalls service
Beadwork: `bedrock-t0k`

A log-process crash with the coordinator still alive restarts the Director with the same epoch. The new full-cluster transaction-history harness observed recovery fail to restore service after this sequence.

`Coordinator.DirectorManagement.handle_director_failure/3` changes only `director` to `:unavailable`, then calls `try_to_start_director/1`. Director startup passes the unchanged `t.epoch`. This differs from coordinator leadership changes, which obtain a new epoch from Raft. See [DirectorManagement at cfaab90a](https://github.com/bedrock-kv/bedrock/blob/cfaab90af3975d11abb208ea2e4a14da8efae133/lib/bedrock/control_plane/coordinator/director_management.ex#L84).

## Observed reproduction

The #239 harness runs a real named-node Cluster supervisor, coordinator/director, sequencer, commit proxy, resolvers, Shale and Olivine through public Repo transactions. After acknowledging a prefix write, it pauses a sentinel append at the synchronous `[:bedrock, :log, :push]` telemetry event. That event occurs after WAL append/fsync and before Shale sends the reply or forwards the append to Demux. The test kills the log while leaving the coordinator alive.

Observed on the draft-fix integration base `1aae9192`:

1. The log and its monitoring Director terminate.
2. Foreman restarts the log.
3. Replacement recovery uses epoch 1 again.
4. The old log retires against an epoch-1 layout.
5. The distributor repeatedly reacquires its lock; the new-epoch/service wait times out.

The source establishes same-epoch reuse. The exact downstream cause of the observed service stall still needs isolation; do not assume that changing one integer locally is a safe fix.

The history harness retains this schedule behind `BEDROCK_HISTORY_SINGLE_LOG_REPRO=1`; its normal bounded schedules explicitly combine coordinator and log loss, or restart the coordinator after a proven partition, to obtain a genuine new Raft epoch. Those schedules do **not** establish automatic director-only recovery. Until #239 lands, its worktree artifacts include `/tmp/bedrock-239-wal.log` and `239-plan-review.md`.

This is distinct from `bedrock-q67.34`, which concerns abandoned recovery-attempt proxies within an epoch.

## Acceptance criteria

- Pin the single-log/live-coordinator failure with a full-core regression containing acknowledged writes and an ambiguous in-flight append.
- Define and implement a consensus-safe fresh recovery generation/epoch and fencing strategy under unchanged coordinator leadership.
- Ensure old transaction components cannot compete with the recovered generation.
- Restore public Repo service, preserve acknowledged writes, and handle the ambiguous tail consistently.
- Test coordinator crash and repeated Director failure around publication, rather than merely incrementing a local counter.

jallum:
Closes #259.

A sole Shale can fail after syncing an in-flight transaction but before replying to its CommitProxy. The client result is ambiguous, the recovery Director and transient transaction components must be replaced, and the acknowledged prefix must remain available. Before the prerequisite fixes, an autonomous Director replacement under the still-live Coordinator reused the same recovery epoch; an old publication could also reinstall stale routing.

This test adds the complete full-cluster history that originally exposed the defect. It is intentionally a test-only integration change: #295 attributes Director publications, #296 allocates and reserves a fresh consensus-backed recovery generation, and #297 fences storage work with the exact `{generation, recovery_id}` grant. The new history proves those contracts compose at the real process, WAL, object-storage, and public Repo boundaries.

The scheduled failure is:

```text
public acknowledged prefix
        ↓
atomic ambiguous tail reaches sole Shale
        ↓
WAL sync ── gate ── reply/demux
        │
        └── kill Shale while Coordinator stays leader
                         ↓
              Director A and transient actors exit
                         ↓
              Director B recovers at greater generation
```

The test captures the exact finalization task at the WAL-sync gate, kills only the sole log, and observes the unknown client result. It requires the old Director, Distributor, sequencer, every proxy and resolver, and that finalizer to be live before the cut and dead afterward. The Coordinator PID, leader node, and Raft term stay unchanged while recovery generation advances from A to B.

Durable recovery state is correlated end to end. The committed B allocation in the retained Raft log, live Coordinator state, synchronized checkpoint, bootstrap reservation, completed bootstrap object, and published layout must all carry B's generation and recovery ID. Every real B Shale and Olivine must be running under that exact grant in memory and in its decoded durable control record.

The ambiguous atomic tail is allowed either legal result: absent, or applied exactly once. Public Repo state must agree with a normal pull from running Shale and with read-only WAL inspection; no version may contain conflicting bytes. The acknowledged prefix must survive, and a fresh post-recovery public transaction must succeed.

The history then attacks B with retired authority. Lower-generation A and equal-generation foreign-ID lock operations target the real Shale and Olivine actors. Structurally valid next-version pushes target Shale. Exact attributed A layout/config messages and legacy forms target the Coordinator, and A's immutable bootstrap reservation token attempts a stale final CAS. All reject while a complete before/after snapshot proves no WAL, materializer, routing, checkpoint, bootstrap, or live worker state changed.

TDD and validation:

- the strengthened test fails on `develop` because the replacement remains generation 1;
- it passes on the stacked implementation with generation 2→3;
- five named-node schedule seeds pass;
- 116 focused notification, generation, publication, Shale, and Olivine tests pass;
- the real publication crash cuts plus this composition test pass 9/9;
- sensitivity mutations detect generation reuse, recovery-ID elision, stale notification acceptance, final-token refresh, and bypass of both Shale owner gates;
- a complete bounded-concurrency distributed run passes 3,075 tests, 158 properties, and 17 doctests;
- format, strict Credo, Dialyzer, and the repository commit hook pass.

Two earlier high-concurrency full runs each hit a different pre-existing load-sensitive test; both exact cases passed immediately in isolation, followed by the clean complete run above.

This draft is stacked on `bedrock-t0k/integration-base`, which contains the reviewed #295, #296, and #297 prerequisite commits.

### DIFF
--- a/test/bedrock/distributed/director_epoch_recovery_test.exs
+++ b/test/bedrock/distributed/director_epoch_recovery_test.exs
@@ -1,0 +1,704 @@
+defmodule Bedrock.Distributed.DirectorEpochRecoveryTest do
+  use ExUnit.Case, async: false
+
+  alias Bedrock.ClusterBootstrap.Publication
+  alias Bedrock.ControlPlane.Config.CoreState
+  alias Bedrock.DataPlane.Log
+  alias Bedrock.DataPlane.Log.Shale.TransactionStreams
+  alias Bedrock.DataPlane.Materializer
+  alias Bedrock.DataPlane.Transaction
+  alias Bedrock.DataPlane.Version
+  alias Bedrock.Service.Foreman
+  alias Bedrock.Service.RecoveryControl
+
+  defmodule Cluster do
+    @moduledoc false
+    use Bedrock.Cluster, otp_app: :bedrock, name: "director_epoch_recovery"
+  end
+
+  defmodule Repo do
+    use Bedrock.Repo, cluster: Cluster
+  end
+
+  @moduletag :distributed
+  @moduletag timeout: 60_000
+
+  test "a log crash under a live coordinator recovers through a fresh fenced generation" do
+    assert Node.alive?(), "run with elixir --sname director_epoch_test -S mix run --no-start"
+
+    root =
+      Path.join(
+        System.tmp_dir!(),
+        "director-epoch-#{System.system_time(:nanosecond)}-#{System.unique_integer([:positive])}"
+      )
+
+    object_root = Path.join(root, "objects")
+    worker_root = Path.join(root, "workers")
+    coordinator_root = Path.join(root, "coordinator")
+    File.mkdir_p!(root)
+
+    previous = Application.get_env(:bedrock, Cluster)
+    previous_storage = Application.get_env(:bedrock, Bedrock.ObjectStorage)
+    backend = {Bedrock.ObjectStorage.LocalFilesystem, root: object_root}
+    Application.put_env(:bedrock, Bedrock.ObjectStorage, backend: backend)
+
+    Application.put_env(:bedrock, Cluster,
+      capabilities: [:coordination, :log, :materializer],
+      durability_mode: :relaxed,
+      path_to_descriptor: Path.join(root, "descriptor"),
+      object_storage: backend,
+      coordinator: [path: coordinator_root],
+      materializer: [path: worker_root, object_storage: backend],
+      log: [path: worker_root, object_storage: backend]
+    )
+
+    start_supervised!({Cluster, []})
+
+    on_exit(fn ->
+      if previous,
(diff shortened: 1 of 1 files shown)


jallum (commented):
lfg

PR #312, opened 2026-09-13
bedrock-jyy: derive release package artifact name in CI

jallum:
Fixes the first-attempt release-package smoke test on `develop`.

The workflow built the current package version but always extracted `bedrock-0.7.0.tar`; it now derives the filename from `mix.exs`.

Verified locally by building 0.7.1, extracting the derived tarball, and confirming the packaged smoke script is present.


jallum (commented):
lfg

PR #314, opened 2026-09-13
Correct 0.7.1 Livebook release notes

jallum:
Fixes the 0.7.1 release notes before retagging.

The note now accurately covers both LocalFilesystem startup fixes (internal lock-file filtering and atomic version-matched writes) and the Livebook re-evaluation fix in #313, which tears down the old distributor placeholder so the replacement does not loop on `lock taken`. The release date is corrected to 2026-09-13.

Documentation-only; no Hex publication.


jallum (commented):
lfg

PR #317, opened 2026-09-13
Hide LocalFilesystem internal files from object listings

jallum:
## Summary

- reserve `.bedrock-` basenames for LocalFilesystem backend metadata
- filter backend-owned files before sorting and applying listing limits
- add a regression proving snapshot discovery ignores both a stale legacy lock and unknown future metadata beside a valid snapshot

## Proof

- regression failed before the implementation: `list("s/0/", limit: 1)` returned `.bedrock-lock`
- focused object-storage and snapshot tests: 67 tests, 0 failures
- patched code read all three snapshots from the untouched failing Livebook object store
- untouched `main` reproduced 7 stale-lock snapshot failures; patched full suite had none of those failures
- full suite: 2,914 tests with one unrelated known flaky commit-proxy telemetry assertion; isolated retry passed
- format, compile, docs, dependency audit, Credo, and Dialyzer pass
- first PR revision passed all four CI jobs

No files are deleted or rewritten, this reservation applies only to LocalFilesystem, and this introduces no NIF.


jallum (commented):
lfg

Issue #319, opened 2026-09-14
Fresh strict cluster startup fails on Link/Foreman order and empty-log recovery


mikehostetler:
## Summary
On Bedrock 0.7.2, a fresh strict three-node cluster cannot complete startup. Two defects appear in sequence:

1. Cluster startup can fail with `:noproc` because Link calls Foreman before Foreman starts.
2. After correcting that startup order locally, Director recovery treats a persisted bootstrap with `logs: %{}` as an old cluster. It enters log recovery and stalls with `:unable_to_meet_log_quorum`, although there are no old logs to recover.

## Reproduction
Use three connected named BEAM nodes. On each node, configure `[:coordination, :log, :materializer]`, a shared LocalFilesystem object store, and strict durability with `desired_logs: 3` and `desired_replication_factor: 3`. Write an initial bootstrap object with the three coordinator nodes and those durability parameters. Start the cluster supervisor on each node.

Observed first: Link registration calls `GenServer.call(foreman_ref, :get_all_running_services, 1000)` while Foreman is not running. The cluster child list starts Link before capability children, which include Foreman. The call exits with `:noproc` (the surrounding `rescue` does not catch exits).

After starting Foreman before Link in a local test, recovery advances but a prior core state with `logs: %{}` goes from TSLValidationPhase to LockingPhase. LogRecoveryPlanningPhase then stalls with `:unable_to_meet_log_quorum`. `CoreState.fresh?/1` already defines a state with no prior logs as fresh, but TSLValidationPhase does not use it.

Source on current `main` (`392417c6`):
- [ClusterSupervisor child order](https://github.com/bedrock-kv/bedrock/blob/392417c671a39a6f271925cc6894e1c10e91a4db/lib/bedrock/internal/cluster_supervisor.ex#L142-L157)
- [Link call to Foreman](https://github.com/bedrock-kv/bedrock/blob/392417c671a39a6f271925cc6894e1c10e91a4db/lib/bedrock/cluster/link/discovery.ex#L195-L210)
- [TSLValidationPhase routing](https://github.com/bedrock-kv/bedrock/blob/392417c671a39a6f271925cc6894e1c10e91a4db/lib/bedrock/control_plane/director/recovery/tsl_validation_phase.ex#L55-L68)
- [Fresh-state definition](https://github.com/bedrock-kv/bedrock/blob/392417c671a39a6f271925cc6894e1c10e91a4db/lib/bedrock/control_plane/config/core_state.ex#L132-L140)

[...]

PR #320, opened 2026-09-14
Start Foreman before Link; route Foreman lookups through the safe call wrapper

jallum:
## Why

`ClusterSupervisor` starts its children in a fixed order: a `DynamicSupervisor`, then `Link`, then whatever capability children the node has (`Foreman`, `Coordinator`, ...). `Link` was started *before* those capability children. That's backwards, because `Link`'s post-init continuation immediately queries the local `Foreman` for its running services, so the coordinator's first view of this node is complete rather than empty:

```elixir
# lib/bedrock/cluster/link/discovery.ex
if :materializer in t.capabilities or :log in t.capabilities do
  case GenServer.call(foreman_ref, :get_all_running_services, 1000) do
```

On a fresh strict cluster this races: `Link` starts, immediately asks `Foreman` for its services — but `Foreman` hasn't started yet, because it's still further down the supervisor's children list. `GenServer.call` to a name nobody has registered exits with `:noproc`. That exit isn't a raised exception, so the `rescue` wrapped around the call doesn't catch it:

```elixir
rescue
  _ -> []
end
```

`rescue` only catches things raised with `raise/1`; a dead `GenServer.call` terminates via `exit/1`, which only `catch :exit, _ -> ...` intercepts. So the "safety net" here was decorative, and the whole `Link` process crashed instead of just returning an empty service list. That's the `:noproc` in #319.

## What changed

Two independent fixes, because either one alone leaves a gap:

1. **`ClusterSupervisor`**: capability children now start before `Link`, so by the time `Link`'s continuation runs, `Foreman` is already registered under its OTP name and the call can't race a nonexistent process.
2. **`Link.Discovery`**: swapped the raw `GenServer.call` + `rescue` for `Bedrock.Service.Foreman.get_all_running_services/2`, the module's own public API. It already goes through `Bedrock.Internal.GenServer.Calls.call/3` — a small wrapper used throughout the project that does the `catch :exit` correctly:

```elixir
catch
  :exit, {:noproc, _} -> {:error, :unavailable}
  :exit, {:timeout, _} -> {:error, :timeout}
  ...
```

So even in some future scenario where the two can still race (a slow node, a crashed Foreman restarting), the call now degrades to `{:error, :unavailable}` — handled by the existing `_ -> []` fallback — instead of taking `Link` down with it.

## How it's tested

Both defects are reproduced as unit tests that fail on `develop` and pass after the fix:

- `cluster_supervisor_test.exs` calls `ClusterSupervisor.init/1` directly (as existing tests in that file already do) and checks that Foreman's normalized child spec appears before Link's in the returned list — a direct assertion on start order, no live supervision tree needed.
- `discovery_test.exs` calls `Discovery.change_coordinator/2` with a log/materializer capability and a deliberately unregistered Foreman name. On `develop` this crashes the test process with the same uncaught `:exit` a real cold boot hits; after the fix it just returns.

Found in review, filed separately as bedrock-0hf rather than folded in here: `ClusterSupervisor.fetch_coordinator/1` has the identical raw-`GenServer.call`-plus-`rescue` bug against `Link`. Same fix, different call site, not part of #319.


jallum (commented):
lfg

PR #321, opened 2026-09-14
Route fresh prior core states to InitializationPhase, not LockingPhase

jallum:
## Why

Recovery keeps two different views of "the previous transaction system": a `CoreState` (durable — just `logs` and `system_materializers`, the minimum needed to recover from) and a `TransactionSystemLayout` (transient — pids, wiped every epoch). `TSLValidationPhase` runs first and type-checks whichever `CoreState` was loaded from the durable bootstrap before anything downstream trusts it. On success it decided where to go next with:

```elixir
def execute(%RecoveryAttempt{} = recovery_attempt, %{prior_core_state: %{} = core_state}) do
  case TSLTypeValidator.validate_type_safety(core_state) do
    :ok ->
      {recovery_attempt, Bedrock.ControlPlane.Director.Recovery.LockingPhase}
```

`%{} = core_state` in a function head doesn't mean "an empty map" — in Elixir it matches *any* map. So this clause fires whether `core_state` is `%{logs: %{"log_1" => [0, 100]}}` (a real prior epoch) or `%{logs: %{}}` (a cluster that has never recovered before, and has nothing to recover from). Both went to `LockingPhase`.

That phase locks the logs named in `prior_core_state` and hands off to `LogRecoveryPlanningPhase`, which computes a majority quorum over them:

```elixir
if locked_count > total_count / 2 do
  ...
else
  {recovery_attempt, {:stalled, :unable_to_meet_log_quorum}}
```

For a fresh cluster, `total_count` is 0. `0 > 0/2` is `false`. Recovery stalls waiting for quorum on logs that were never there in the first place — the bug in #319.

The fix already exists in the codebase and is just unused here: `CoreState.fresh?/1` correctly treats an absent or empty-logs `CoreState` as fresh, and `MaterializerBootstrapPhase` already asks it this exact question for the system shard. `TSLValidationPhase` was the one phase that skipped the check.

## What changed

`TSLValidationPhase` now asks `CoreState.fresh?/1` after validation succeeds, and only locks the old system when there's an old system to lock:

```elixir
defp next_phase(core_state) do
  if CoreState.fresh?(core_state) do
    InitializationPhase
  else
    LockingPhase
  end
end
```

`InitializationPhase` seeds a fresh layout (log vacancies, default shard split) and goes straight to recruitment — no locking, no quorum math, because there's nothing from a previous epoch to preserve.

Along the way, adding this call surfaced a real but previously invisible type mismatch: `TSLTypeValidator.validate_type_safety/1` was specced to take a full `TransactionSystemLayout.t()`, but `TSLValidationPhase` has only ever called it with the narrower `CoreState.t()` — that's why the validator's `resolvers` check was already documented as "vacuous against this record." Dialyzer only caught it once `next_phase/1` also typed its argument as `CoreState.t()`, at which point the two specs contradicted each other for the same variable.

The one function had two real callers with two genuinely different shapes: `TSLValidationPhase` (a `CoreState`) and `assert_type_safety!/1` — used with a real `TransactionSystemLayout` in `topology_phase.ex` — which delegates to the same function internally. Rather than paper over that with a union type (`TransactionSystemLayout.t() | CoreState.t()`, which would let either caller pass either shape without dialyzer objecting), split it into two precisely-typed entry points sharing one private, unspecced implementation:

```elixir
@spec validate_layout_type_safety(TransactionSystemLayout.t()) :: :ok | {:error, term()}
def validate_layout_type_safety(%{} = tsl), do: validate_fields(tsl)

@spec validate_core_state_type_safety(CoreState.t()) :: :ok | {:error, term()}
def validate_core_state_type_safety(%{} = core_state), do: validate_fields(core_state)
```

`TSLValidationPhase` now calls `validate_core_state_type_safety/1` directly. Same fix rippled one hop further: `trace_recovery_tsl_validation_failed/2`, the telemetry call in the failure branch, had the identical stale `TransactionSystemLayout.t()` typing for what is, at its one call site, always a `CoreState` — narrowed that too, and dropped the now-permanently-empty "Resolvers" line from its corruption log message (a `CoreState` never has one).

## How it's tested

`tsl_validation_phase_test.exs` had a test literally named "transitions to LockingPhase when TSL validation succeeds" that passed `logs: %{}` and asserted `LockingPhase` — it was pinning the bug as expected behavior. Split into two cases: fresh (`logs: %{}`) now asserts `InitializationPhase`; a new non-fresh case (`logs: %{"log_1" => [1, 2]}`) asserts `LockingPhase` is unchanged for real prior epochs.

That ripples into `recovery_test.exs` and `service_locking_integration_test.exs`, whose default test fixtures represent a fresh cluster and were asserting the old stall. With the fix, those recoveries now progress further and stall for more honest reasons — e.g. `{:insufficient_nodes, 2, 1}` (recruiting `desired_logs: 2` from a single test node) instead of `:unable_to_meet_log_quorum` over logs that never existed. Two tests were renamed since their old names ("...still stalls at log quorum") no longer described what they test.


jallum (commented):
lfg

PR #322, opened 2026-09-14
bedrock-80f: prove a real multi-node cluster can be driven from ExUnit

jallum:
Phase 0 of a chaos-testing harness. This is a spike: it answers one falsifiable question and deliberately builds nothing on top of the answer.

## Why

We want a harness that stands up a few nodes, runs a workload, and randomly kills things — processes often, whole nodes less so — to find out what breaks. Everything in that plan rests on an assumption nobody had tested: that a real multi-node Bedrock cluster can be started and driven from ExUnit at all.

That assumption was worth doubting. There is no multi-node test facility anywhere in the repo — no `:peer`, no `:slave`, no `LocalCluster`. Every existing "integration" test runs in a single BEAM against mock clusters. And a single BEAM *cannot* host a multi-coordinator cluster: `Descriptor.coordinator_nodes` is a list of `node()`, capabilities are per-node config, and `ClusterSupervisor` refuses to start on `:nonode@nohost`. Real distribution is the floor, not an upgrade. Meanwhile `bedrock_ex`'s `ISSUES.md` documents cross-node worker init failures and orphaned workers from doing this by hand.

So: prove it works, or find out precisely what blocks it, before building a harness on sand.

## What

Three `:peer` nodes form one cluster, a transaction commits on the first node, and the value reads back from the second. Green across repeated runs in ~1.3s, with no orphaned BEAM processes and no leftover temp directories.

The test is tagged `:chaos` and excluded by default, following the existing `:distributed` and `:s3` gating in `test_helper.exs`. Run it with `BEDROCK_CHAOS=1 mix test --include chaos`. The default suite is unaffected.

Reading back from a *different* node than the one that committed is the point of the assertion — a value visible only on its writer would mean the commit never reached a shared transaction system.

## How

Three things turned out to be non-obvious, and each is documented where it bites:

**Readiness needs two stages.** `ClusterSupervisor.child_spec/1` silently falls back to a single-node descriptor when it cannot read the descriptor file, so three nodes will cheerfully come up as three separate one-node clusters that each look perfectly healthy. Stage one therefore requires every node to agree on the full coordinator set; only then does stage two wait for a transaction system layout, which exists only after recovery completes. "All nodes healthy" was true in the broken case and told us nothing. Filed as `bedrock-y3c`, since an operator with a bad descriptor path hits the same trap and gets invisible writes.

**Object storage has to be shared.** Cluster bootstrap state lives in object storage, so giving each node a private store yields N nodes that each bootstrap their own cluster and never converge — presenting as a timeout, not an error. One `LocalFilesystem` root serves all three, which also keeps MinIO out of this entirely. `:object_storage` is set both at top level (coordinator/director) and per capability (foreman), because they read it from different places.

**`:erpc.call` cannot start a supervisor.** It runs work in a process that exits the moment it has a result and signals that result *as its exit reason*, so a tree started with `Supervisor.start_link` from inside is linked to an already-dying process and goes down with it. An unlinked owner spawned on the peer holds the tree instead.

One caveat worth naming: that owner is a parked `spawn`, so nothing supervises the cluster root. That is fine here — `:peer.stop` tears down the whole VM — but when fault injection needs to observe the root dying, the owner should become a monitored process rather than a sleeping one.

Workload code lives in compiled `test/support` rather than the `.exs` test file, because peers inherit the primary's code paths but `.exs` files are only ever evaluated in the primary VM; a closure built in a test file has no module on the peer.

## Scope

No fault injection, workload generation, oracles, or scheduling — those come next, stacked on this branch. The only thing kept beyond the passing test is the peer-startup helper it needed.

Also filed from this work: `bedrock-kxb` (`:resolution` appears in the public capability typespec but `module_for_capability/1` raises on it).


mikehostetler:
## Peer-helper coordination for Jido recovery regressions

I prepared two small Bedrock-only cases while checking Jido Agent persistence: loss of a log/system-cache node, and loss of the elected leader after followers booted before the bootstrap existed. Each starts four peers with three logs / RF3, acknowledges one binary checkpoint, and checks exact restoration plus the next revision. They pass on the local #326 + #254 adaptation + #334 combination at seeds 0 and 7.

For local verification I used a small compiled peer helper with an unlinked GenServer owner, peer monitors, layout telemetry, state barriers, and a bounded read transaction for readiness. It stops the peer-local test-only Mox application before connecting peers to avoid its global-server collision. There are no timing sleeps.

I do not want to submit a second peer framework. Before offering the regressions on #254/#326, I would like to reuse this PR's helper and add only the readiness/fault hooks it needs. The existing #328 merge-order note also applies to our current-code configuration; we have not integrated #328 yet.

All peers run on one host with shared LocalFilesystem; these are node-loss application checks, not a partition or independent-host durability proof. The helper and new cases remain local and uncommitted.

PR #323, opened 2026-09-14
bedrock-3t9: seed a fresh cluster's sizing from the node config the durability check validates

jallum:
Tracks bedrock-3t9.

## Why

Strict durability mode is supposed to refuse to start a cluster that can't meet the durability profile — three logs, replication factor three. The guides tell you to satisfy it with node config:

```elixir
config :my_app, MyCluster,
  durability_mode: :strict,
  durability: [desired_logs: 3, desired_replication_factor: 3]
```

`ClusterSupervisor.init/1` checks exactly that node config, and it passes. But nothing ever *applied* it: a fresh cluster's parameters come from `Config.new/1` → `Parameters.new/1`, which hardcodes `desired_logs: 1, desired_replication_factor: 1`. In a three-node harness cluster configured as above, every node started cleanly in strict mode and the cluster ran with a single log. `Bedrock.Durability.profile(MyCluster)` — which prefers the live cluster config — reported `:failed`, and nothing acted on it. The guardrail was checking a number the cluster ignored, and there was no configuration path to more than one log at all.

## What changed

When a coordinator starts the first recovery of a fresh cluster (no bootstrap record in object storage), `DirectorManagement.maybe_put_default_config/1` now seeds `desired_logs` and `desired_replication_factor` from the same evaluation the startup check performs — `Profile.evaluate(node_config).checks.*.actual` — instead of leaving `Config.new/1`'s hardcoded ones. Those parameters are persisted in the first cluster bootstrap, so later boots and recoveries keep them.

Existing clusters are untouched: once a bootstrap exists its parameters win, and editing node config afterwards does not resize the cluster (that needs an operator write path — bedrock-q67.51). `guides/durability-profile.md` now says so, and that `durability:` must be identical across coordinator nodes, since the seed comes from whichever coordinator leads the first recovery.

## How it works

Reading the checks' `actual` values rather than re-reading node config makes "what was validated" and "what is seeded" the same computation, including the fallback to 1 for missing or malformed values. The evaluation is done on the keyword list, not the cluster module: given a module, `Profile.evaluate/1` would `fetch_config/0` from the very coordinator that is running this code.

Worth knowing while reviewing: `desired_replication_factor` sizes nothing today — the replication factor the cluster actually runs with is `max(1, map_size(logs))` (topology phase), so `desired_logs` is what really buys durability. Seeding both keeps the cluster-level profile consistent with the node-level check; making the replication factor its own knob is part of bedrock-q67.50.

The regression test (`director_monitoring_test.exs`) drives `try_to_start_director/1` with no config and node config asking for 3/3, and asserts the seeded parameters are 3/3; before the change it failed with `desired_logs: 1, desired_replication_factor: 1`.


jallum (commented):
lfg

PR #326, opened 2026-09-14
bedrock-w10: recover from the durable bootstrap, not the coordinator's copy

jallum:
Tracks bedrock-w10.

## Why

A coordinator that wins raft leadership starts a Director, and the Director's recovery begins from a *prior core state* — which logs the previous generation wrote to. It was being handed the coordinator's in-memory copy. That copy is loaded once, when the coordinator process boots, and afterwards only updated by `{:notify_transaction_system_layout, ...}`, which a Director sends to its own coordinator — the leader. Followers never hear it.

So when a follower becomes leader, its Director recovers from whatever that follower saw at boot:

- If the follower booted before the cluster's first bootstrap existed, that's `nil`. `CoreStateValidationPhase` reads `nil` as a brand-new cluster, and recovery does exactly what it would for one. In a three-node cluster on develop, killing the leader (which held the only log) produced, within 90ms on the new leader:

  ```
  Initializing a brand new system
  Bedrock recruitment: replaced log candidates that failed to lock ["vpktuo3v"] with fresh workers ["jas222qp"]
  Fresh cluster detected, using default shard layout
  Recovery completed in 90ms!
  ```

  and the bootstrap in object storage was rewritten to name only the new, empty log. The cluster stayed up and returned `nil` for every previously committed key.

- If the follower booted at an older epoch, it names logs that a later recovery already retired, and its Director waits forever for a quorum of logs that will never exist again.

FDB doesn't have this shape: cluster recovery reads the durable core state (`cstate.read()` in `clusterRecoveryCore`) before it does anything else, rather than trusting any process's memory.

## What changed

The Director reads the bootstrap from object storage itself, once, when its recovery starts (`Recovery.try_to_recover/1` in `:starting`), through a new `PersistencePhase.read_bootstrap/1` — so it reads exactly the object that phase later overwrites.

- A record is found → its core state is the prior state.
- `:not_found` → `nil`. Only a genuinely absent record means a fresh cluster.
- Any other error (object storage unreachable, unreadable record) → recovery doesn't start. The Director stays `:starting` and retries every second — FDB's `MASTER_SPIN_DELAY` — instead of treating "couldn't read" as "fresh", and instead of crashing into an immediate restart loop.
- No object storage configured → there is no durable record, and the coordinator's copy is still all there is (unchanged behavior).

The coordinator no longer loads a core state at boot; its `prior_core_state` now only holds what its own Directors report, which is the no-object-storage fallback above. `trace_director_launch` stops printing a prior-log count, since a follower's count was exactly the misleading number in this bug.

## How it works

```elixir
defp read_prior_core_state(t) do
  case PersistencePhase.read_bootstrap(t.cluster) do
    {:ok, bootstrap} -> {:ok, CoreState.from_bootstrap(bootstrap)}
    {:error, :not_found} -> {:ok, nil}
    {:error, :no_object_storage} -> {:ok, t.prior_core_state}
    {:error, _reason} = error -> error
  end
end
```

The read happens once per Director, before any phase consumes the prior state; retries of a *stalled* recovery don't re-read, matching FDB's fixed `prevDBState` per recovery. It can't move backwards: a Director only notifies its coordinator after its bootstrap write succeeds, so the durable record is always at least as new as any in-memory copy. This also covers the leader's own Director crashing between persisting the bootstrap and notifying.

Tests (`durable_prior_state_test.exs`) cover the three shapes: a `nil` prior state while the bootstrap names logs (asserts no vacancy seeding — the wipe), a stale prior state yielding to the bootstrap's current logs, and an unreadable bootstrap leaving the Director `:starting` and retrying.

Known gaps, tracked separately: clusters without a shared object store still have no durable core state for a new leader to read (bedrock-yzg), and the persistence write isn't fenced on the version recovery started from (bedrock-1d8).


mikehostetler:
## Jido Agent persistence verification — 2026-09-15

I tested your implementation at `8d0bb980c046858373ddf0c36d5bbe2198c69125`. I will use this Director-time durable read rather than submit our older coordinator-side refresh. The distinction between `:not_found` and a failed read is important.

The new Bedrock-only peer regression has this small application schedule:

1. Boot four peers fresh, before any bootstrap exists, using three logs / RF3.
2. Commit one versioned binary checkpoint and read its exact bytes from another node.
3. Kill the elected coordinator leader.
4. Wait for completed recovery through layout events, state barriers, and a real transaction.
5. Check unchanged cluster identity, exact revision-one bytes, a successful revision-two commit, stale-revision rejection, and exact revision-two bytes.

This case and a separate log/system-cache-node loss case pass at seeds 0 and 7. No load or timing sleep is needed. The native regression uses ordinary KV transactions, not Jido-specific Bedrock behavior.

**The availability result includes** a current-code adaptation of #254 and the locked-log guard in #334. It must not be read as a result from this PR alone, or as closure of the Link/election/ownership gaps already noted here. Your nil/stale/unreadable-bootstrap tests are included in the passing 98-test focused selection.

Jido's read-only readiness probe now permits normal Bedrock retries within its existing ten-second deadline because layout publication precedes asynchronous data-cache adoption. Jido checkpoint writes still use `retry_limit: 0`; this is not a new upstream write-retry feature.

The peers run on one host with shared LocalFilesystem. These tests do not establish independent-host failure, power-loss durability, or partition safety.

I can supply a test-only follow-up using the #322 peer helper after we agree on its integration base. No competing implementation PR is planned.

PR #328, opened 2026-09-14
bedrock-1cp: resolve the cluster's object storage backend in one place

jallum:
Tracks bedrock-1cp.

## Why

A Bedrock cluster keeps three kinds of durable state in object storage: the **cluster bootstrap** (written by the director, read by the coordinator on cold boot), **log chunks** (written by workers), and **materializer snapshots**. Until now each was located by a different rule:

| Reader/writer | Where it looked |
|---|---|
| `Coordinator.Server` (bootstrap read) | top-level `object_storage:`, else `LocalFilesystem` under the first of `coordinator/log/storage/materializer/coordination` paths |
| `PersistencePhase` (bootstrap write) | top-level `object_storage:`, else the first of `log/storage/materializer/coordination` — no `coordinator` |
| `ClusterSupervisor` → Foreman → workers (chunks) | `object_storage:` inside the merged `log:`/`materializer:`/`worker:` sections, else that section's path |
| Olivine (snapshots) | app config `config :bedrock, Bedrock.ObjectStorage, backend:`, default `System.tmp_dir!()/bedrock_objects` |

So a configuration could quietly split a cluster's durable state. Setting up a MinIO-backed three-node cluster by putting S3 in `log:` (the natural place if you're reading the Foreman) sent chunks to S3 while every node read and wrote its bootstrap on local disk. That left each node with its own cold-boot anchor, so the cluster as a whole had none. Setting only the top level gave the reverse. Even on one node, differing coordinator and log paths made the director write the bootstrap where the coordinator never looks, so every restart came up fresh. And the README's app-config S3 setting only ever reached snapshots.

## What changed

One resolver, `Bedrock.ObjectStorage.Config.cluster_backend/1`, now answers "where does this cluster's durable state live?" for every site:

1. `object_storage:` at the top level of the node config, normalized so the `:s3` shorthand works (`object_storage: :s3, s3: [...]`);
2. otherwise a `LocalFilesystem` at `<first role path>/object_storage`, using the coordinator's existing order (`coordinator, log, storage, materializer, coordination, worker`). Wherever the old coordinator could find a bootstrap, the new code finds the same one.

The coordinator, `PersistencePhase`, and `ClusterSupervisor` call it (their private copies are deleted). The Foreman hands the result to every worker, and Olivine builds its snapshot handle from that value instead of reading app config.

Two placements that used to be silently honored in part are now refused at startup:

- `object_storage:` inside a capability section: `":object_storage belongs at the top level of the node config, not in the :log section"`.
- `config :bedrock, Bedrock.ObjectStorage, backend: ...`: honoring it now would point an upgraded coordinator at a store holding no bootstrap, and it would cold-boot as a fresh cluster. The error says so and asks for a deliberate relocation.

`Config.backend/0`, `Config.bootstrap_key/0,1` and `ClusterBootstrap.Discovery.discover/1` lose their last callers and are deleted. README, `guides/object-storage-s3.md` and `guides/durability-foundation.md` show the node-config form. The CHANGELOG lists every storage-location move, including the one where chunks move for configs whose Foreman path differed from the coordinator path.

## How it works

A multi-node S3/MinIO node config is now one key, identical on every node:

```elixir
config :my_app, MyApp.Cluster,
  capabilities: [:coordination, :log, :materializer],
  coordinator: [path: "/var/lib/bedrock"],
  log: [path: "/var/lib/bedrock"],
  materializer: [path: "/var/lib/bedrock"],
  object_storage: {Bedrock.ObjectStorage.S3, [bucket: "bedrock", config: [host: "minio.internal", port: 9000, ...]]}
```

`test/bedrock/object_storage/resolution_test.exs` drives each previously-splitting configuration through the three real sites and asserts they agree: the Foreman's backend, where the director's bootstrap lands, and where the coordinator cold-boots from. Before the change these failed with the Foreman on `<node>/object_storage` instead of the shared root, and with the coordinator getting `{:error, :not_found}` for a bootstrap the director had written elsewhere.

Note for merging: #326 (bedrock-w10) adds `PersistencePhase.read_bootstrap/1` on top of the helper this PR deletes. Whichever lands second should route that read through `cluster_backend/1`.


mikehostetler:
## Separate S3 write-body defect and integration follow-up

While checking Jido's native snapshot upload, I found an independent S3 iodata body defect and opened #335. ObjectStorage/Snapshot accept iodata, but all three S3 PUT paths pass a list to ExAws's body encoding. Binary bundles can raise `Jason.EncodeError` before HTTP.

The small proposed fix converts iodata to a binary at the S3 boundary, without changing storage selection, conditional headers, or retries. Five service-free tests use the real ExAws signing/request path and check exact bytes, content length, payload hash, conditional writes, and Snapshot bundles.

Our current-code real MinIO profile passes 29/29 twice after that fix, including native upload and cold reconstruction without the manual upload workaround. **#328 is not in that tested combination.** The recovery checks also use #326, a #254 adaptation, and #334.

When we test an accepted combination with this PR, we will remove the old capability-level/app-level storage placement and use its top-level cluster config. The new #326 bootstrap read must use the shared resolver as already noted in this PR. We will not attribute the current passes to the new resolver before rerunning that configuration.

